PR #10, opened 2022-07-27
fix bad merge

pivnicek:
@ErikZigo I cancelled the CI because it would interfere with the other run.  This is just a fix of the tagging pipeline that I broke when fixing merge confilcts


pivnicek:
fixed elsewhere

PR #101, opened 2023-05-31
remove confusing warning

odinuv:
the preserve flag doesn't make any sense for local staging (because it is not persistent), so it is neither true nor false - it is is ignored 
https://keboola.atlassian.net/browse/PST-856


pepamartinec (approved):
lgtm

PR #136, opened 2023-07-21
fix handling of processed tags in OM

odinuv:
follow up to https://github.com/keboola/platform-libraries/pull/135

handles the BC break created in IM getFiles method


ErikZigo (approved):
Přihodil jsem test. Přijde mi to ok.

PR #137, opened 2023-07-24
[api-bundle] Add Vault service URL

pepamartinec:
https://keboola.atlassian.net/browse/SOX-133

Pridani Vault service


MiroCillik (approved):
Jasne 👍

PR #185, opened 2023-11-09
Almost remove StorageApiToken class

odinuv:
https://keboola.atlassian.net/browse/GCP-462

depends on https://github.com/keboola/storage-api-php-client-branch-wrapper/pull/23


libs/api-bundle/composer.json
```diff
@@ -14,7 +14,7 @@
         "keboola/api-error-control": "^4.1",
         "keboola/coding-standard": "^14.0",
         "keboola/kbc-manage-api-php-client": "^7.0",
-        "keboola/storage-api-php-client-branch-wrapper": "^4.0",
+        "keboola/storage-api-php-client-branch-wrapper": "dev-odin-GCP-462",
```
ErikZigo, line 17: nezapomenout prehodit na stable verzi

PR #310, opened 2024-08-14
[PST-1988] convert default/length int/bool values to string

ondrajodas:
https://keboola.atlassian.net/browse/PST-1988


ErikZigo (approved):
lgtm

PR #311, opened 2024-08-15
[PST-2000] Default value ignored when adding/changing column

ondrajodas:
https://keboola.atlassian.net/browse/PST-2000


ErikZigo (approved):
lgtm

PR #318, opened 2024-09-09
[input-mapping] Table/Result::getTables() return type phpdoc

ErikZigo:
Part of https://keboola.atlassian.net/browse/PST-1689

Abych v job runneru nemel phpstan errory

```
Parameter #1 $tableInfo of static method App\Helper\TableInfoConverter::convertTableInfoToTableResult() expects Keboola\InputMapping\Table\Result\TableInfo,  
         mixed given. 
```

### DIFF
--- a/libs/input-mapping/src/Table/Result.php
+++ b/libs/input-mapping/src/Table/Result.php
@@ -22,6 +22,9 @@ class Result
         $this->tables[] = $table;
     }
 
+    /**
+     * @return Generator<TableInfo>
+     */
     public function getTables(): Generator
     {
         foreach ($this->tables as $table) {


libs/input-mapping/src/Table/Result.php
```diff
@@ -22,6 +22,9 @@ public function addTable(TableInfo $table): void
         $this->tables[] = $table;
     }
 
+    /**
+     * @return Generator<TableInfo>
```
ErikZigo, line 26: https://github.com/keboola/platform-libraries/blob/fc314aea144197df8e1f0eee53acf764175edd71/libs/output-mapping/src/Table/Result.php#L24

AdamVyborny (approved):
👍🏻

PR #321, opened 2024-09-16
fix[service-client]: Fix switched namespace of billing & buffer services

pepamartinec:
Nahodou jsem narazil na to, ze mame prohozeny namespace u `billing` a `buffer` servis (`billing` bezi v `default` NS, `buffer` ve svym NS).

image

image


romantmb (approved):
😄 👍

PR #337, opened 2024-12-02
Don't change length if hasn't been changed

ondrajodas:
https://keboola.atlassian.net/browse/PST-2296


ErikZigo (approved):
lgtm 👍 snad tam neceka dalsi prekvapeni

PR #357, opened 2025-01-17
[output-mapping] Disable native-types feature if new-native-types is set

ondrajodas:
https://keboola.atlassian.net/browse/PST-2394


ErikZigo (approved):
lgtm

PR #373, opened 2025-02-19
DMD-67 disable clone load for external buckets, allow view load for tables with external bucksts

zajca:
DMD-67


odinuv (approved):
jop, to by mělo fungovat

PR #412, opened 2025-08-19
Add Query Service to Service Client

ujovlado:
Changes:

- Add Query Service to Service Client

---

### DIFF
--- a/libs/service-client/src/Service.php
+++ b/libs/service-client/src/Service.php
@@ -16,6 +16,7 @@ enum Service
     case IMPORT;
     case NOTIFICATION;
     case OAUTH;
+    case QUERY;
     case QUEUE;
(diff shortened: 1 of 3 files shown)


libs/service-client/src/Service.php
```diff
@@ -60,6 +62,7 @@ public function getInternalServiceName(): string
             self::IMPORT => 'sapi-importer.default',
             self::NOTIFICATION => 'notification-api.default',
             self::OAUTH => 'oauth-api.default',
+            self::QUERY => 'query-service-api.query-service',
```
MiroCillik, line 65: Why is one some services the suffix `default` and on some there is a name of the service like `query-service`? :)

pepamartinec (approved):
LGTM

PR #489, opened 2026-04-17
[AJDA-2302] fix(output-mapping): fix result.json lookup path

ondrajodas:
## Problem

`loadCustomVariables` was looking for `result.json` at the metadata storage root (e.g. `/data/result.json`), but the file is placed in the output directory (e.g. `/data/out/result.json`).

The metadata storage path for local staging equals the data dir (`/data`), without the `out/` prefix. The `sourcePathPrefix` (e.g. `out/tables`) points to the tables subdirectory, so `result.json` lives one level up from that.

## Fix

Use `dirname(rtrim($configuration->getSourcePathPrefix(), '/')))` combined with `metadataStorage->getPath()` to correctly resolve the output directory containing `result.json`.

Before:
```php
Path::join($strategy->getMetadataStorage()->getPath(), 'result.json')
// => /data/result.json (wrong)
```

After:
```php
Path::join($strategy->getMetadataStorage()->getPath(), dirname(rtrim($configuration->getSourcePathPrefix(), '/')), 'result.json')
// => /data/out/result.json (correct)
```

### DIFF
--- a/libs/output-mapping/src/TableLoader.php
+++ b/libs/output-mapping/src/TableLoader.php
@@ -170,7 +170,11 @@ class TableLoader
 
         $tableQueue = new LoadTableQueue($this->clientWrapper, $this->logger, $loadTableTasks);
         $tableQueue->start();
-        $tableQueue->loadCustomVariables(Path::join($strategy->getMetadataStorage()->getPath(), 'result.json'));
+        $tableQueue->loadCustomVariables(Path::join(
+            $strategy->getMetadataStorage()->getPath(),
+            dirname(rtrim($configuration->getSourcePathPrefix(), '/')),
+            'result.json',
+        ));
 
         return $tableQueue;
     }


ondrajodas:
```
# Release Notes
Fixed `result.json` lookup path in `output-mapping`. Custom variables defined in `result.json` were silently ignored because the file was searched at the wrong path (`/data/result.json` instead of `/data/out/result.json`).

No customer communication needed — this is an internal bug fix. Components relying on custom variables from `result.json` will now work correctly without any changes on their side.

# Impact Analysis
Affects all jobs using local staging that write `result.json` with custom variables. Previously the file was never found, so `getCustomVariables()` always returned an empty array. No breaking change — behavior is now correct.

# Change Type
Bug fix

# Justification
AJDA-2302 — `result.json` was placed at `out/result.json` relative to the data dir, but `loadCustomVariables` was looking at the data dir root. The fix uses `dirname(sourcePathPrefix)` to resolve the correct output directory.

# Deployment Plan
No deployment steps required. Library change only — consumers pick up the fix on next dependency update.

# Rollback Plan
Revert the single-line change in `TableLoader.php` and release a patch version.

# Post-Release Support Plan
Verify that jobs using `result.json` for custom variable output now correctly propagate those variables. No special monitoring needed beyond standard job output validation.
```

ondrajodas:
@odinuv pls, jedna drobnost, zkoušel jsem tam existenci result.json na špatný cestě

odinuv (approved):

odinuv:
Ještěže jsme se o tom bavili no a měl jsem to blbě, no :)

PR #490, opened 2026-04-24
feat(sandboxes-service-api-client): add type filter and type field to AppsApiClient

odinuv:
https://linear.app/keboola/issue/AJDA-1942

## Description
Adds a `$types` array parameter to `AppsApiClient::listApps()` that sends `type[]=value` query params to the `/apps` endpoint, enabling server-side filtering by app type (e.g. `python`, `r`, `streamlit`). Also maps the nullable `type` field from the `ExistingApp` API response into the `App` model.

## Release Notes

**Change Type**
New feature — adds type filtering support and exposes the `type` field on the `App` model.

**Justification**
The CLI needs to list only Python/R sandbox apps when replacing the legacy sandboxes-api with the new apps API. Without this filter, callers would receive all app types and have to filter client-side.

**Plans for Customer Communication**
No customer communication needed — this is an internal API client library used by other Keboola services.

**Impact Analysis**
Low risk. Additive change only: new optional parameter defaults to `[]` (no filtering), and the new `type` field on `App` is nullable. No breaking changes.

**Deployment Plan**
Standard CI/CD continuous deployment across all stacks.

**Rollback Plan**
Standard git revert if issues detected post-deployment.

**Post-Release Support Plan**
No follow-up actions needed.

### DIFF
--- a/libs/sandboxes-service-api-client/src/Apps/AppsApiClient.php
+++ b/libs/sandboxes-service-api-client/src/Apps/AppsApiClient.php
@@ -19,11 +19,12 @@ class AppsApiClient
     }
 
     /**
-     * @param int|null $offset
-     * @param int|null $limit
+     * @param int|null      $offset
+     * @param int|null      $limit
+     * @param list<string>  $types Filter by app type (e.g. 'python', 'r', 'streamlit')
      * @return array<App>
      */
-    public function listApps(?int $offset = null, ?int $limit = null): array
+    public function listApps(?int $offset = null, ?int $limit = null, array $types = []): array
     {
         $queryParams = [];
         if ($offset !== null) {
@@ -31,6 +32,9 @@ class AppsApiClient
         }
         if ($limit !== null) {
             $queryParams['limit'] = (string) $limit;
+        }
+        foreach ($types as $type) {
+            $queryParams['type'][] = $type;
         }
 
         $uri = '/apps';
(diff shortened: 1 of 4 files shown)


libs/sandboxes-service-api-client/src/Apps/AppsApiClient.php
```diff
@@ -19,19 +19,31 @@ public function __construct(ApiClientConfiguration $configuration)
...
             $queryParams['offset'] = (string) $offset;
         }
         if ($limit !== null) {
             $queryParams['limit'] = (string) $limit;
         }
+        foreach ($types as $type) {
+            $queryParams['type'][] = $type;
+        }
+        if ($listAll) {
+            $queryParams['listAll'] = '1';
```
ondrajodas: listAll funguje?
https://github.com/keboola/sandboxes-service/blob/1689c3e322ea89f99dfc3030e4222c65ec2d0c6f/src/Controller/Apps/ListAppsQuery.php

ondrajodas (commented):
příde mi že to nemůže fungovat

libs/sandboxes-service-api-client/src/Apps/AppsApiClient.php
odinuv: Nefunguje, máš naprostou pravdu. To je nějaká halucinace (moje, ne AI)

odinuv: Je to pryč

ondrajodas (approved):
super, ale ještě se to bude muset teda upravit v tom cličku

PR #502, opened 2026-05-20
PAT-1795 Drop default loginType from staging-provider workspace options

pepamartinec:
## Release Notes
https://linear.app/keboola/issue/PAT-1795/switch-job-snowflake-workspace-to-always-use-logintype-snowflake

Stops `WorkspaceProvider::prepareNewWorkspaceOptions()` from injecting `loginType: default` when the caller passes `null`. The `loginType` key is now only present in the Storage API workspace-create options when an explicit value is supplied. This prepares the API for the follow-up `job-queue-job-configuration` change that will start passing `snowflake-service-keypair` unconditionally for Snowflake staging.

`WorkspaceProvider` becomes the canonical place that decides what to send for `loginType`: nothing if the caller didn't ask, otherwise exactly what was asked. Connection treats absence of the key as `default`, so the behavior remains identical for current callers.

None.

## Impact analysis
No end-user impact. The previous behavior (passing `loginType: default`) and the new behavior (omitting the key) are equivalent on the Storage API side — `default` is what Connection assumes when the key is absent.

## Change type
Refactoring

## Justification
Prepares `staging-provider` API for PAT-1795: switching Snowflake job workspaces to keypair auth unconditionally.

## Deployment
Merge & automatic deploy.

## Rollback plan
Revert of this PR.

## Post release support plan
None.

### DIFF
--- a/libs/staging-provider/src/Workspace/WorkspaceProvider.php
+++ b/libs/staging-provider/src/Workspace/WorkspaceProvider.php
@@ -11,7 +11,6 @@ use Keboola\StagingProvider\Staging\StagingType;
 use Keboola\StagingProvider\Workspace\Configuration\NewWorkspaceConfig;
 use Keboola\StorageApi\ClientException;
 use Keboola\StorageApi\Components;
-use Keboola\StorageApi\WorkspaceLoginType;
 use Keboola\StorageApi\Workspaces;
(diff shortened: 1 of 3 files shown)


odinuv (approved):
👍

PR #503, opened 2026-05-22
PAT-1791 git-service-api-client: Kubernetes SA / JWT auth strategies

pepamartinec:
## Release Notes
https://linear.app/keboola/issue/PAT-1791

Adds an \`AuthenticationStrategyInterface\` to \`keboola/git-service-api-client\` mirroring the one shipped in \`keboola/kbc-manage-api-php-client\` master, so in-cluster callers — initially \`sandboxes-service\` — can authenticate to git-service with a projected Kubernetes ServiceAccount JWT instead of a static Manage.

Three strategies are included: \`ManageApiTokenAuthenticationStrategy\` (existing \`X-KBC-ManageApiToken\` path; bare-string tokens still work via the constructor for backwards compatibility), \`StaticJwtAuthenticationStrategy\` (sends \`X-Kubernetes-Authorization: Bearer <jwt>\`), and \`KubernetesServiceAccountTokenAuthenticationStrategy\` (same header but reads the JWT from a projected-token file on every request so kubelet-rotated tokens are picked up automatically). \`ApiClient\` / \`GitServiceApiClient\` constructors accept either a non-empty string (legacy) or an \`AuthenticationStrategyInterface\`; the Guzzle \`mapRequest\` middleware resolves headers per request instead of capturing them once.

None.

## Impact analysis
No end-user impact. Existing string-token callers compile and behave identically (the \`string\` constructor path wraps the value in \`ManageApiTokenAuthenticationStrategy\`). The new strategies are opt-in.

## Change type
New feature

## Justification
Lets sandboxes-service authenticate to git-service with a projected SA token instead of a long-lived static manage-token.

## Deployment
Merge & automatic deploy.

## Rollback plan
Revert of this PR.

## Post release support plan
None.

### DIFF
--- a/libs/git-service-api-client/README.md
+++ b/libs/git-service-api-client/README.md
@@ -1,9 +1,42 @@
 # Git Service API Client
 
 ## Installation
+
 ```bash
 composer require keboola/git-service-api-client
 ```
+
+## Usage
+
+```php
+use Keboola\GitServiceApiClient\ApiClientConfiguration;
+use Keboola\GitServiceApiClient\Auth\KeboolaServiceAccountAuth;
+use Keboola\GitServiceApiClient\Auth\ManageApiTokenAuth;
+use Keboola\GitServiceApiClient\GitServiceApiClient;
+
+// Default: projected Kubernetes ServiceAccount token from
+// /var/run/secrets/connection.keboola.com/serviceaccount/token, re-read on
+// every request so kubelet-rotated tokens are picked up automatically.
+$client = new GitServiceApiClient('https://git-service.example.com');
+
+// Manage API token (legacy)
+$client = new GitServiceApiClient(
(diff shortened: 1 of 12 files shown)


zajca (approved):

romanbracinik (approved):

pepamartinec:
Some final touches:
* added `KeboolaServiceAccountAuth` which is original `KubernetesServiceAccountAuth` but preconfigured to expect the token at `/var/run/secrets/connection.keboola.com/serviceaccount/token` (we don't use default K8S tokens, but custom issued tokens with `keboola-connection` as audience)
* dropped `KubernetesServiceAccountAuth` and `KubernetesTokenAuth` as they're not gonna be used anywhere (we can always add them back later if nessecary)

PR #506, opened 2026-06-01
[AJDA-2322] Migrate CI from Azure Pipelines to GitHub Actions

ondrajodas:
Migrates the platform-libraries monorepo CI from Azure Pipelines to GitHub Actions, with a full cutover (Azure files deleted in this PR).

## Change & dependency detection
- New native PHP CLI `bin/ci/affected-libraries.php` (+ `AffectedLibrariesResolver`, unit tests) replacing `bin/ci-find-changes.sh`.
- Reads every `libs/*/composer.json`, builds the `*@dev` path-repo reverse dependency graph, computes the transitive closure, and outputs the affected libraries (changed ∪ dependents) as JSON.
- **Fallback to "all"** when shared infrastructure changes (`.github/`, `bin/`, `Dockerfile`, `docker-compose*`, `.dockerignore`, root `composer.json`/`composer.lock`), on first commit / zero-SHA / force-push, or on any error. (This PR itself touches `.github/` + `bin/`, so it runs all 22 libraries.)
- Runs natively on the runner via `shivammathur/setup-php` (no Docker) so the `detect-changes` gate stays fast. Its own checks: `docker compose run --rm dev82 bash -c 'cd bin/ci && composer ci'`.

## Workflows
- **`.github/workflows/ci.yml`** — orchestrator (PRs and pushes to `main`): `detect-changes` → 22 conditional per-library test jobs → `tests-result` barrier (single required status check) → 22 conditional `publish` jobs (push to `main` only).
- **`.github/workflows/release.yml`** — 22 tag-gated release jobs (`refs/tags/<lib>/*`), ported from `azure-pipelines.tags.yml`. Tag → split/publish only, **no tests** (the code was already tested on merge to `main`).
- **`.github/workflows/lib-*.yml` × 22** — reusable `workflow_call` workflows; config stays with each package.
  - Special cases preserved 1:1: `input-mapping`/`output-mapping` multi-suite (cs / AWS / Azure / BigQuery); `k8s-client` & `messenger-bundle` Terraform provisioning; `logging-bundle` Symfony 6.4/7.2 matrix.
- **`.github/actions/split-library`** — composite action wrapping `bin/split-repo.sh`, minting a short-lived GitHub App installation token at runtime (via `actions/create-github-app-token`, scoped to the target repo) from `SPLIT_APP_ID` + `SPLIT_APP_PRIVATE_KEY` and pushing over HTTPS.

## CI configuration split (variables vs secrets)
- Non-sensitive values read via **`vars.*`** (repository variables): Storage API URLs, `HOSTNAME_SUFFIX_GCP`, `OUTPUT_MAPPING__BIGQUERY_STORAGE_API_URL`, AWS access key IDs.
- Sensitive values read via **`secrets.*`** (repository secrets, `secrets: inherit`): all Storage API tokens, Terraform secret access keys, `SPLIT_APP_PRIVATE_KEY`.
- Full list documented in `CLAUDE.md`.

## Concurrency & UX
- Every per-library workflow serializes via a `<lib>-lock` concurrency group (`cancel-in-progress: false`) — the same library never runs twice in parallel against shared storage resources (the second run queues).
- All workflow **steps** have explicit names (no bare `Run …` in the UI).
- Multi-suite **jobs** have readable display names (e.g. `cs` → `Code style & static analysis`, suites labelled by cloud and test group; logging-bundle → `Symfony 6.4` / `Symfony 7.2`).

## Dependency fix
- `libs/logging-bundle/composer.json`: set `config.policy.advisories.block=false`. Composer 2.10 (2026-05-28) blocks installing advisory-affected versions by default; a fresh CI image build pulls 2.10, which made the Symfony 7.2 resolution fail (symfony/yaml, monolog-bridge, cache whole ranges currently flagged). Azure passed only on an older cached Composer. Scoped to logging-bundle's own CI; not inherited by consumers. Verified locally: `validate` + `install` + `composer ci` all green for Symfony 7.2.

## Cutover (deleted)
`azure-pipelines.yml`, `azure-pipelines.tags.yml`, `azure-pipelines/`, all `libs/*/azure-pipelines.tests.yml`, `bin/ci-find-changes.sh`. Kept `bin/split-repo.sh` and `bin/adopt-repo.sh`.

## Required from repo admin (out-of-band)
- Create/install a **GitHub App** with push rights to all 22 standalone repos; expose its App ID as the `SPLIT_APP_ID` variable and its private key as the `SPLIT_APP_PRIVATE_KEY` secret. The composite action mints a short-lived, repo-scoped installation token at runtime. (Until set, tag/push publishing fails at auth — tests are unaffected.)
- Provision the repository variables and secrets listed in `CLAUDE.md`.
- Set `tests-result` as the single required status check in branch protection; remove the old Azure check.

## Verification
- `bin/ci` `composer ci`: phpcs clean, PHPStan level max `[OK] No errors`, PHPUnit green.
- `actionlint` exit 0 across all workflow/action files.

Linear: AJDA-2322


ondrajodas:
```
# Release Notes
CI for the platform-libraries monorepo moves from Azure Pipelines to GitHub Actions (hard cutover — Azure removed in the same change). Pull requests test only the changed libraries plus their transitive `*@dev` dependents (with a fallback to "all" when shared infra changes); pushes to `main` test affected libraries and then mirror/publish them to their standalone repos; tag pushes (`<lib>/x.y.z`) publish a single library (no tests). Each library serializes via a concurrency lock so it never runs twice in parallel against shared storage. No change to the libraries' published artifacts.

Internal-only change (CI/build infrastructure). No external customer communication needed. The team owning this repo must know the required status check changed from the Azure check to the GitHub Actions `tests-result` job, and that CI config is now provisioned as repository variables + secrets.

# Impact Analysis
Affected: the build/test/publish pipeline of all 22 libraries. Risks: (1) change-detection correctness — covered by unit tests and a fallback-to-all on any uncertainty; (2) publish/release now depend on a GitHub App installation token minted at runtime from `SPLIT_APP_ID` + `SPLIT_APP_PRIVATE_KEY` instead of Azure service connections; (3) Terraform flows (k8s-client, messenger-bundle) and the logging-bundle Symfony matrix were ported 1:1 and need a live run to confirm; (4) a logging-bundle Composer config change (`policy.advisories.block=false`) was required because Composer 2.10 now blocks advisory-affected versions on install — scoped to logging-bundle CI only. No application/runtime code changed.

# Change Type
CI/CD infrastructure migration (config change) + one dependency-policy fix in logging-bundle. No production application code.

# Justification
Consolidate CI on GitHub Actions and remove the Azure Pipelines dependency (Linear AJDA-2322, related PAT-1346 / PAT-1791). Enables config-with-package reusable workflows and native change/dependency detection.

# Deployment Plan
1. Repo admin creates/installs a GitHub App with push rights to all 22 standalone repos and sets its App ID as the `SPLIT_APP_ID` variable and its private key as the `SPLIT_APP_PRIVATE_KEY` secret.
2. Provision the repository variables (URLs, host suffix, AWS access key IDs) and secrets (Storage API tokens, Terraform secret keys, `SPLIT_APP_PRIVATE_KEY`) listed in `CLAUDE.md`.
3. Update branch protection: set `tests-result` as the single required status check; remove the old Azure check.
4. Merge PR (hard cutover — Azure files deleted in the same change).
5. Validate the first push-to-main mirrors affected libraries, and a test tag (`<lib>/x.y.z`) releases via `release.yml`.

# Rollback Plan
Revert the PR. This restores `azure-pipelines.yml`, the per-lib `*.tests.yml` templates, `azure-pipelines.tags.yml`, and `bin/ci-find-changes.sh` intact, and removes the GitHub Actions workflows and the logging-bundle config change. Re-point branch protection back to the Azure status check.

# Post-Release Support Plan
Watch the first few PR runs and the first push-to-main: confirm `detect-changes` selects the correct libraries, the `tests-result` barrier gates correctly, publish jobs mirror to the right standalone repos with correct tag prefixes, and the first tag release works. Edge cases to monitor: zero-SHA/first-commit/force-push fallback-to-all, GHA cache behaviour across the 4 PHP base images, the Terraform provisioning for k8s-client/messenger-bundle, and whether the Composer advisory situation (logging-bundle / other symfony-heavy libs) needs a broader fix (repo-wide policy or pinned Composer version).
```

pepamartinec (commented):
Jak jsem psal na Slacku, taky jsem par tydnu zpatky zkousel neco podobnyho, ale zaparkoval to nekde pres secretama a uz se k tomu benratil. Pro inspiraci jsem to ted pushnul https://github.com/keboola/platform-libraries/tree/refs/heads/pepa/gha

Par veci, co je potreba zmenit:

## 🔴 1. We're dropping per-branch library publishing (developer-workflow regression)

Today's Azure pipeline (`trigger: '*'`, `pr: none`, `publishLibraries` gated only on `succeeded('testsResults')` with **no branch check**) splits every affected library to its standalone `keboola/<lib>` repo on **every green build, on any branch**. Combined with `split-repo.sh`'s `git clone --mirror` + `git push --mirror` (which preserves all branch refs — `filter-repo` only rewrites tags), this means pushing a dev branch makes the matching branch appear in the standalone repo, so we can:

```
composer require keboola/input-mapping:dev-my-feature-branch
```

…in a dependent project and test an unmerged library change end-to-end.

## 🟠 2. Publish credentials should be least-privilege, not a shared long-lived secret

Every `publish-*`/`release-*` job uses `secrets.LIBS_SPLIT_TOKEN` — one long-lived token with push rights to **all** standalone repos. The composite action's `token` input is literally documented as *"GitHub App installation token with push rights to the target repo,"* so the intent is already there.

**Suggestion:** mint a short-lived token per run, scoped to the single target repo, with `actions/create-github-app-token` (the action already accepts a `token` input, so it's a drop-in):

```yaml
- uses: actions/create-github-app-token@v1
  id: app-token
  with:
    app-id:       ${{ vars.SPLIT_APP_ID }}
    private-key:  ${{ secrets.SPLIT_APP_PRIVATE_KEY }}
    owner:        keboola
    repositories: ${{ <target-repo-name> }}   # scope to just this lib's repo
- uses: ./.github/actions/split-library
  with:
    token: ${{ steps.app-token.outputs.token }}
    ...
```

Auditable, auto-expiring, and a leaked token can't touch 21 other repos. (Note: the repo already has `SPLIT_APP_ID` / `SPLIT_APP_PRIVATE_KEY` configured, so the App appears to exist — we'd just switch from the static token to minting installation tokens from it.)

Jeste poznamecka - nebylo by lepsi to obecne resit pres nejakou GitHub appku nez staticky `LIBS_SPLIT_TOKEN`?

## 🟠 3. The affected-libraries resolver is never tested or linted in CI

`bin/ci/` ships `phpunit.xml.dist`, `phpstan.neon`, `phpcs.xml`, and `tests/AffectedLibrariesResolverTest.php`, but `detect-changes` only runs `composer install` and then executes the resolver. **The component that decides which libraries get tested is itself untested in the pipeline** — if it silently under-detects (e.g. a dependency-graph edge case), affected libs get skipped and we'd never know.

Poznamecka: moc se mi nelibi tohle resit v PHP, preferoval bych na to Bash skripty, sedi mi to vic na ten ukol, ale to je jen osobni preference.

## 🟡 4. ~66 near-identical job blocks — consider matrix + a reusable workflow

There are 22 `lib-*.yml` files, 22 `publish-*` jobs in `ci.yml`, and 22 `release-*` jobs in `release.yml`. Most libraries are identical (`composer ci` in `dev-<lib>`); the per-lib differences are just data (PHP version, compose service, target-repo name).

**Suggestion:** drive the simple libraries from a `matrix` feeding **one** reusable `lib-common.yml` (and similarly for publish/release), keeping bespoke workflows only for the genuinely special ones (`input-mapping`, `output-mapping`, `k8s-client`, `messenger-bundle`, `logging-bundle`). Cuts the per-lib boilerplate by ~3×, and adding a library becomes a one-line data change instead of three new job blocks — less drift, smaller diffs to review.

Poznamecka: mrkni na tu moji branch, tam jsme to tak nadesignovali. Nova libka se pak smrskne na par radek misto komple novyho worflow (pokud nepotrebuje nejaky speciality).

## 🟡 5. No orchestrator-level concurrency → superseded runs aren't cancelled

`ci.yml` has no top-level `concurrency`, so pushing twice to an open PR runs the whole fan-out twice in parallel.

**Suggestion:**
```yaml
concurrency:
  group: ci-${{ github.ref }}
  cancel-in-progress: true
```
Keep the per-library `cancel-in-progress: false` locks as-is — those correctly serialize against shared Storage API projects.

## 🟢 6. `release.yml` triggers on all tags

`on: push: tags: ['*']` spins up the workflow and evaluates all 22 `startsWith(...)` job conditions for *any* tag (e.g. a repo-level `v*`). Minor, but `tags: ['*/*']` matches only the `<lib>/<version>` shape we actually use.

---

Co bych jeste zvazil:
* ma tu nejaky duvod/vyhodu to triggerovat pres `pull_request`, ne plain `push`?
* potrebujeme poustet CI na `main`?
* zachoval si tu chovani "pokud nenajdu konkretni zmenenou lib, pustim vse" - pohaval jsme si myslenkou, jestli to potrebujeme? pokud se nezmenila zadna lib, tak se v podstate nic nemeni/nereleasuje

ondrajodas:
kouknu se na to... jen k tvým komentářům:
> ma tu nejaky duvod/vyhodu to triggerovat pres pull_request, ne plain push

asi nijak zvláštní důvod to nemá

> potrebujeme poustet CI na main

ano potřebujeme... měly by se pouštět jen dotčené libky + právě dependency libky které závisí na té změněné... to kvůli tomu že tu jedeme přes @dev závislosti tak aby si měl jistotu že to jinde nerozbiješ

> zachoval si tu chovani "pokud nenajdu konkretni zmenenou lib, pustim vse" - pohaval jsme si myslenkou, jestli to potrebujeme? pokud se nezmenila zadna lib, tak se v podstate nic nemeni/nereleasuje

no o tom jsem taky přemýšlel no, pokud si řekneme že to nepotřebujeme tak to změním... za mě ale lepší takto... změníš Dockerfile, docker-compose, cokoliv a může nastat problém v nějaký libce...proto jsem se rozhodl spíš takto

ondrajodas:
@pepamartinec 

ad 1. máš pravdu
ad 2. no však to je přes github appku.. možná nemáš pullnutý poslední commit z dnešního rána?
ad 3. a mě se zas líbilo to mít takto no :-D 🤷‍♂️ 
ad 4. no zase.. mě se líbilo že vidíš workflow `lib-*`...otevřeš ho a nemusíš o tom přemýšlet....
ad 5. 👍 
ad 6. 👍

ondrajodas:
fixed:
ad 1, 5 a 6 fixnutá

MiroCillik (commented):
## Review — Azure Pipelines → GitHub Actions migration

Great migration overall: the change-detection rewrite into a tested PHP tool, the per-library reusable workflows, the runtime-minted scoped App token, and the careful "push only the build branch / single tag" split logic are all well thought through and documented in `CLAUDE.md`. The 1:1 port of the multi-suite jobs and concurrency mutexes preserves the existing behaviour faithfully.

A few things worth addressing before merge — one I'd consider blocking (the required-check gap), the rest are hardening / doc fixes. Details inline.

### Summary
- **🔴 `tests-result` can be green while nothing was tested** — if `detect-changes` fails, every lib job is *skipped* (not failed), and the barrier treats `skipped` as OK. Since `tests-result` is the single required check, a broken detector = mergeable PR with zero tests run. (inline on ci.yml)
- **🟠 `secrets: inherit` over-shares** — every per-library test job (which builds & runs third-party `composer` code) receives *all* repo secrets, including `SPLIT_APP_PRIVATE_KEY` and every Storage token. Consider passing only what each workflow needs. (inline)
- **🟡 `publish` is not gated to `main`** — it runs on every branch push (intentional per `CLAUDE.md`, but `README.md` says "On push to `main`"). Reconcile the docs, or add the gate if that was the intent. (inline)
- **🟡 Redundant image builds** — each of input-mapping's ~9 jobs (and output-mapping's) rebuilds the same image from scratch with no shared cache; the old Azure flow built once. Not blocking, but slower/costlier. (inline)
- Nit: `bin/split-repo.sh` usage synopsis line lists only `[push-branch]` though it now takes a 6th `[push-tag]` arg.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,222 @@
...
+      - php-test-utils
+      - query-service-api-client
+      - sandboxes-service-api-client
+      - service-client
+      - settle
+      - slicer
+      - staging-provider
+      - sync-actions-api-php-client
+      - vault-api-client
+    if: always()
```
MiroCillik, line 204: 🔴 **Required check can pass without running any tests.**

`tests-result` only `needs:` the 22 lib jobs — not `detect-changes`. If `detect-changes` fails (git error, PHP fatal, etc.), every lib job is *skipped* because its `needs: [detect-changes]` is unmet. This barrier counts `skipped` as acceptable, so it reports **success** — and since this is the single required status check, the PR becomes mergeable with **zero libraries tested**.

The `catch`-all in `affected-libraries.php` mitigates most failures (it falls back to `allDirs()`), but anything that crashes the runner step *before* the script's own fallback (setup-php failure, `composer install` failure in `bin/ci`, OOM) reopens the hole.

Suggest: add `detect-changes` to `tests-result`'s `needs:` and fail if its result isn't `success`, e.g. include it in the same `toJSON(needs)` check.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,222 @@
...
+        run: |
+          LIBS=$(php bin/ci/affected-libraries.php '${{ steps.base.outputs.sha }}')
+          echo "libs=${LIBS}" >> "$GITHUB_OUTPUT"
+          echo "Affected: ${LIBS}"
+
+  api-bundle:
+    needs: [detect-changes]
+    if: contains(needs.detect-changes.outputs.libs, '"api-bundle"')
+    uses: ./.github/workflows/lib-api-bundle.yml
+    secrets: inherit
```
MiroCillik, line 44: 🟠 **`secrets: inherit` hands every secret to every test job.**

Each `lib-*.yml` runs `docker compose run … 'composer install && composer ci'`, i.e. it pulls and executes third-party dependency code with network access. With `secrets: inherit`, that job's scope includes **all** repository secrets — every Storage token plus `SPLIT_APP_PRIVATE_KEY` (the publish App key) — even though a given lib only references one or two. A compromised transitive dependency in any single library could exfiltrate the publish key and all tokens.

The publish job reads `SPLIT_APP_PRIVATE_KEY` directly in `ci.yml` (line 222), so test workflows never need it. Consider declaring `secrets:` explicitly per reusable workflow (only the tokens that workflow uses) instead of `inherit`. More verbose, but it contains the blast radius.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,222 @@
...
+    steps:
+      - name: Verify no test job failed
+        run: |
+          RESULTS='${{ toJSON(needs) }}'
+          echo "$RESULTS"
+          echo "$RESULTS" | python3 -c "import sys, json; r=json.load(sys.stdin); bad=[k for k,v in r.items() if v['result'] not in ('success','skipped')]; sys.exit(1) if bad else print('All test jobs passed or were skipped')"
+
+  publish:
+    needs: [detect-changes, tests-result]
+    if: github.event_name == 'push' && needs.tests-result.result == 'success' && needs.detect-changes.outputs.libs != '[]'
```
MiroCillik: 🟡 **`publish` runs on every branch, not just `main`.**

The only trigger is `on: push: branches: ['**']`, so `github.event_name == 'push'` is always true — this job splits/publishes on *any* branch push. `CLAUDE.md` documents this as intentional (so `dev-<branch>` is `composer require`-able), but `README.md:87` says "On push to `main`, affected libraries are split/published". Please reconcile: either fix the README, or add `github.ref == 'refs/heads/main'` if main-only was actually intended.

README.md
```diff
@@ -79,6 +79,22 @@ docker compose run --rm dev-input-mapping bash -c "vendor/bin/phpunit tests/Func
 docker compose run --rm dev-input-mapping bash -c "vendor/bin/phpunit --filter testTableManifest tests/Functional/DownloadTablesTest.php"
 ```
 
+## Continuous Integration
+
+CI runs on **GitHub Actions** (`.github/workflows/`).
+
+- `ci.yml` (PRs and pushes to `main`) detects affected libraries via `bin/ci/affected-libraries.php` — the changed libraries plus their transitive `*@dev` dependents — and runs only the matching per-library workflows (`lib-<lib>.yml`). Infra-wide changes test all libraries. The `tests-result` job is the single required status check.
+- On push to `main`, affected libraries are split/published to their standalone repos.
```
MiroCillik: 🟡 This says publishing happens "On push to `main`", but `ci.yml`'s `publish` job has no `main` gate — it runs on every branch push (consistent with `CLAUDE.md`). Update this line to "On every branch push" (or add the gate in `ci.yml`).

.github/workflows/lib-input-mapping.yml
```diff
@@ -0,0 +1,223 @@
...
+  cancel-in-progress: false
+
+jobs:
+  cs:
+    name: Code style & static analysis
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v6
```
MiroCillik, line 16: 🟡 **Each suite rebuilds the image from scratch.** All ~9 jobs in this workflow (and output-mapping's) independently run `docker compose build dev-input-mapping` with no shared layer cache — the old Azure flow built once and shared the artifact across stages. Not blocking, but it multiplies build time/cost across every affected run. Consider a `build` job that pushes to a registry cache + `cache-from`, or GHA `docker/build-push-action` caching, and have suites pull.

.github/workflows/ci.yml
ondrajodas: Good catch. Added `detect-changes` to `tests-result`'s `needs:`, so its result now goes through the same `toJSON(needs)` barrier — if the gate fails (and every lib job gets skipped) the required check now fails instead of reporting success. Also documented the rationale in CLAUDE.md.

.github/workflows/ci.yml
ondrajodas: Thanks. Keeping `secrets: inherit` for now. The runtime blast radius is narrower than it looks: GitHub Actions doesn't auto-inject secrets into the job environment — each `lib-*.yml` only maps the tokens it references onto `env:`, so `SPLIT_APP_PRIVATE_KEY` (never referenced in any test workflow) isn't present where the third-party `composer install`/`composer ci` code runs and can't be exfiltrated by a compromised dependency. `inherit` is also a documented design choice (CLAUDE.md). Explicit per-workflow `secrets:` across all 22 workflows is a worthwhile least-privilege hardening, but we'll track it as a separate follow-up rather than block this migration.

.github/workflows/ci.yml
ondrajodas: Intentional and documented in CLAUDE.md — publishing on every branch is what makes `composer require keboola/<lib>:dev-<branch>` work in dependents. The README was the thing out of sync; updated it to "On every branch push" (and fixed the trigger description above it). Keeping `ci.yml` as-is.

README.md
ondrajodas: Fixed — updated to "On every branch push" to match the intentional behavior, and also corrected the line above it ("PRs and pushes to main" → every branch push; there's no `pull_request` trigger).

.github/workflows/lib-input-mapping.yml
ondrajodas: Agreed it's not blocking. A shared build cache (a `build` job pushing to a registry/GHA cache with `cache-from`, suites pulling) is a good optimization but out of scope for this migration PR — tracking it as a separate follow-up.

MiroCillik (approved):
OK. All main issues were resolved.
Please create task for the follow up with image caching.

PR #509, opened 2026-06-04
refactor(input-mapping)!: unify table strategies on two-phase load contract (AJDA-2841)

odinuv:
## Link to issue

https://linear.app/keboola/issue/AJDA-2841/input-mapping-refactor-before-removal-of-the-load-decider

## Description

Unifies all five table-input strategies (Local, S3, ABS, workspace Snowflake/BigQuery) on a two-phase load contract, replacing the legacy `downloadTable()`/`handleExports()` template method:

- **Phase 1** `prepareAndExecuteTableLoads(tables, preserve): TableLoadQueueInterface` — plans the loads and submits the Storage jobs (workspace: one unified `workspaceLoad` job with `loadType` per input; file strategies: one export job per table via the new split `TableExporter` API).
- **Phase 2** `waitForTableLoadCompletion(queue): Result` — agnostic core in `AbstractStrategy` (await, metrics, Result/state assembly) + per-strategy `materializeTableLoads()` hook (manifests, S3/ABS credential enrichment, Local file download).
- `Reader::downloadTables()` is now the composition of the two phases; new `Reader::waitForTableLoadCompletion()` lets outside callers finish a queue without knowing strategy classes — queues carry their creating strategy class + destination for that.
- The legacy workspace path (2 jobs: `workspaceLoadClone` + `workspaceLoad`, `useView` flag) is deleted; the clone/copy race workaround is obsolete since Connection handles everything in a single job.
- Legacy `use_view` config flag is normalized to `load_type=VIEW` at the configuration layer and honored by the load decider; it no longer leaks past the normalizer.
- Removed unused `Reader::getParentRunId()`.

## BC breaks (release as a new major/minor)

- `StrategyInterface`: `downloadTable()`/`handleExports()`/`downloadTables()` removed; two-phase methods added.
- `Reader::prepareAndExecuteTableLoads()` return type widened `WorkspaceLoadQueue` → `TableLoadQueueInterface` (runtime object for workspace stagings is still `WorkspaceLoadQueue`; callers reading `->jobs` need an `instanceof` narrow or migrate to `getJobIds()`/`getAllTables()`).
- `WorkspaceLoadQueue`/`TableExportQueue` constructors gained `strategyClass` + `destination` params.
- `InputTableOptions::isUseView()` removed → use `getLoadType() === 'VIEW'`; `use_view` no longer appears in parsed configuration output (still accepted as input; explicit `load_type` wins on conflict).
- `Reader::getParentRunId()` removed (no known consumers).

## Behavior changes

- Workspace input mapping submits **1** Storage job instead of 2 (metrics count 2 → 1, log `Processed 1 workspace exports.`).
- Empty table list + `preserve=false` on a workspace now queues a clean job (workspace is wiped) where legacy was a no-op — consistent with `preserve=false` semantics; persistent workspaces use `preserve=true` and are unaffected.
- Local staging writes manifests only after a successful load (legacy wrote them before export, leaving orphans on failure); `verifyToken()` once per batch instead of per table.

## Justification

See the linked Linear issue — removes the duplicated legacy/new load paths and unblocks the follow-up removal of the client-side load decider.

None (internal library; component-visible behavior changes are listed above and benign).

## Impact Analysis

All consumers of workspace input mapping switch to the single-job load format on upgrade. Manifests/metrics/Result contract preserved (covered by the functional suites). Single-tenant stacks: no special handling; the unified `workspaceLoad` API is GA. The empty+preserve=false drift is the one observable semantic change — flagged above.

## Deployment Plan

Release a tagged version (BC-breaking bump). Requires `keboola/storage-api-client` ^18.9.0 (split `TableExporter` API) — already pinned.

## Rollback Plan

Revert this PR.

## Post-Release Support Plan

None.


libs/input-mapping/src/Table/Strategy/AbstractWorkspaceStrategy.php
```diff
@@ -254,34 +180,6 @@ public function prepareAndExecuteTableLoads(array $tables, bool $preserve): Work
         return $this->executeTableLoadsToWorkspace($plan);
     }
 
-    /**
-     * @deprecated Used only by handleExports to preserve legacy behavior
-     */
-    private function buildCopyInputLegacy(
```
ondrajodas, line 260: 🎉

ondrajodas (approved):
jako proč ne no, bude se to muset pořádně vyzkoušet  než to nasadíme

PR #510, opened 2026-06-07
fix(messenger-bundle): AJDA-2866 apply default REST timeout to all GPS transports

odinuv:
### Link to issue

https://linear.app/keboola/issue/AJDA-2866/notification-service-consumer-can-hang-indefinitely-on-blocking

### Description

A Pub/Sub consumer using the REST connection has no client-side timeout on the long-poll `pull()` — a half-open TCP connection blocks `curl_exec()` forever, and neither `--time-limit` nor SIGTERM can interrupt it (16.5h outage on `com-keboola-gcp-us-east4`). The bundle already injects `client_config.restOptions` timeouts for the transports it creates itself, but app-defined `gps://` transports (like notification-service's `messages`) got nothing.

This adds `GpsTransportFactoryDecorator`, which decorates the petitpress `GpsTransportFactory` (registered unconditionally, with `decoration_on_invalid: ignore` for apps without `GpsMessengerBundle`), injecting `restOptions: {timeout: 120, connect_timeout: 10}` into every GPS transport. 120s exceeds the ~90s Pub/Sub server-side long-poll hold, so healthy polls are never cut short.

Non-obvious decisions:
- The decorator backs off when `client_config` is set explicitly — checking **both** `$options` and the DSN query, because `GpsConfigurationResolver` merges them with `$options` taking precedence, so an injected value would silently override a DSN-configured one.
- It backs off on the whole `client_config`, not just `restOptions`, to avoid shadowing an app-configured `requestTimeout` (Google's `RequestWrapper` gives `restOptions` precedence).
- `KeboolaMessengerExtension` GCP defaults now reference the decorator's constant instead of duplicating the literals.

### Justification

See the linked issue — on timeout the worker now exits with `TransportException` and the container restarts, making stuck consumers self-healing.

### Plans for Customer Communication

None — internal infrastructure change.

### Impact Analysis

Affects all apps using this bundle on GCP once they upgrade. Behavior change only for `gps://` transports with no explicit `client_config`: requests now fail after 120s instead of hanging forever. A timeout mid-pull surfaces as a `TransportException` that crashes the worker — which is the intended recovery path. Apps with explicit `client_config` are untouched. No feature flag.

### Deployment Plan

Release a tagged version.

### Rollback Plan

Revert this PR.

### Post-Release Support Plan

None.

### DIFF
--- a/libs/messenger-bundle/README.md
+++ b/libs/messenger-bundle/README.md
@@ -30,6 +30,11 @@ keboola_messenger:
   connection_audit_log_queue_dsn: "%env(CONNECTION_AUDIT_LOG_QUEUE_DSN)%"
 ```
 
+### Default timeouts for Google Pub/Sub transports
+When installed, the bundle gives every `gps://` transport in the application a default REST client timeout
+(`restOptions.timeout: 120`, `connect_timeout: 10`) — without it, a consumer can hang forever on a dead connection.
+Setting `client_config` explicitly (in transport options or the DSN query) disables the defaults.
+
 ## Development
 Prerequisites:
 * `aws` CLI with configured `Keboola-Dev-Platform-Services-AWSAdministratorAccess` profile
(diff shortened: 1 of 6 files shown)


pepamartinec (approved):
👌

PR #507, opened 2026-06-02
PAT-1860 DMD-1475 feat(api-bundle): exchange Connection programmatic tokens for Storage tokens

zajca:
## What & why

`api-bundle` can now accept the new Connection **programmatic bearer tokens** (`kbc_at_*` access
tokens and `kbc_pat_*` personal access tokens) on endpoints that today require a legacy Storage
token.

Inspired by the [`auth-bridge-proxy` RFC](https://github.com/keboola/go-monorepo/blob/main/docs/rfcs/2026-05-18-auth-bridge-proxy.md),
but **decentralized**: instead of a central Go proxy, each service performs the exchange itself.
It calls Connection's internal resolver endpoint (added in keboola/connection#7403, extended in
keboola/connection#7604) via the Manage API client's `Client::resolveStorageToken()`,
authenticating with its **own projected Kubernetes ServiceAccount JWT**
(`X-Kubernetes-Authorization`), hands over the user's token (`X-Subject-Token`), and receives the
matching **legacy Storage token together with its full token detail** (the same payload as
`GET /v2/storage/tokens/verify`). The `StorageApiToken` is built directly from that single
response — **no follow-up Storage verify call** (keboola/connection#7604 made the second
round-trip unnecessary).

## How it works

```
Authorization: Bearer kbc_at_*, X-KBC-ProjectId: 123
        │
        ▼  POST /manage/internal/auth-bridge/resolve-storage-token
   Connection (validates subject token + project scope, decrypts)
        │
        ▼  { storageToken, tokenDetail, ... }
   => new StorageApiToken(tokenDetail, storageToken)  ->  #[CurrentUser] StorageApiToken
```

## Integration

Transparent and always on: existing `#[StorageApiTokenAuth]` controllers also accept programmatic
tokens — no controller change, no configuration switch. Exchange triggers only for
`Authorization: Bearer kbc_at_*/kbc_pat_*` (+ `X-KBC-ProjectId`); legacy `X-StorageApi-Token`
traffic is unaffected.

## Notable design points

- The resolver call lives in the Manage API client (`keboola/kbc-manage-api-php-client` `^v10.2`,
  `Client::resolveStorageToken()`), a direct dependency of the bundle.
- ServiceAccount token is read from the projected file **per request** (kubelet rotation).
- No result caching in v1 (revoked tokens stop working immediately); a short-lived cache can be
  added later, per the RFC.
- Token material (subject token, resolved Storage token + detail, SA JWT) is never logged or put
[...]

PR #518, opened 2026-06-12
DMD-1475 Add bootCleanClient() test helper for stubbing auth on a clean container

pepamartinec:
## Release Notes
https://linear.app/keboola/issue/DMD-1475

TL;DR uprava test traitu aby fungoval i s novou zavislosti na `ManageApiClientFactory`. V testech appek je pak potreba mechanicka uprava viz. https://github.com/keboola/sandboxes-service/pull/612

Adds `AuthenticatorTestTrait::bootCleanClient()` to `api-bundle`. Since the token-exchange work (#507), a `#[StorageApiTokenAuth]` request initializes `ManageApiClientFactory` (it backs the exchange resolver), and Symfony's test container refuses to replace an already-initialized service. So a functional test that performs a Storage-token request and *then* calls `setupFakeManageApiToken()` on the same container (the common `disableReboot()` pattern) fails with *"service is already initialized, you cannot replace it."* `bootCleanClient()` boots a fresh kernel/client, giving a clean container in which `setupFake*Token()` can stub the auth services after a request has run.

The trait now requires a `WebTestCase` host — it declares abstract `bootKernel()`/`getClient()`, and `bootCleanClient()` is `static` to make its global (kernel-level) effect explicit. The existing `setupFake*Token()` helpers are unchanged. api-bundle's own trait test is converted to a `WebTestCase`, which required wiring the minimal kernel so it actually boots.

Key changes:
* `bootCleanClient()` + abstract `bootKernel()`/`getClient()` in `AuthenticatorTestTrait`.
* Test kernel boots for real: registers `ValinorBundle`, keeps `ManageApiClientFactory` + the resolver client public (no firewall to retain them otherwise), sets `app_name`/`secret`/`HOSTNAME_SUFFIX`.
* Adds `symfony/browser-kit` + `symfony/dom-crawler` dev deps.

None.

## Impact analysis
No runtime/end-user impact — test-support code only (`src/Test`, `tests/`) plus two dev dependencies. Consumers that adopt `bootCleanClient()` must extend Symfony's `WebTestCase` (the trait now declares abstract `bootKernel()`/`getClient()`); existing `setupFake*Token()` usage is unaffected.

## Change type
Improvement

## Justification
Let consuming services stub Manage/Storage auth after a request without the "service already initialized" error introduced by the token-exchange wiring (DMD-1475).

## Deployment
Merge & automatic deploy.

## Rollback plan
Revert of this PR.

## Post release support plan
None.

### DIFF
--- a/libs/api-bundle/README.md
+++ b/libs/api-bundle/README.md
@@ -148,31 +148,86 @@ which the bundle requires directly, so no extra installation is needed.
 ## Testing controllers
 
 `Keboola\ApiBundle\Test\AuthenticatorTestTrait` stubs the authenticators in functional
-(`KernelTestCase`) tests so guarded controllers can be exercised without reaching real
-Storage/Manage APIs:
+(`WebTestCase`) tests so guarded controllers can be exercised without reaching real
+Storage/Manage APIs. It provides four helpers:
+
+| Helper | Stubs auth for | Returns |
+| --- | --- | --- |
+| `setupFakeStorageApiToken(tokenString?, projectId, features, adminId?)` | `#[StorageApiTokenAuth]` via legacy `X-StorageApi-Token` | `StorageApiToken` |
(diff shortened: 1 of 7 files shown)


zajca (approved):
Imho cajk

PR #522, opened 2026-06-16
PAT-1866 Per-client exception types and HTTP context on the base client

pepamartinec:
## Release Notes
https://linear.app/keboola/issue/PAT-1866/unified-api-client-base

> [!NOTE]
> Diky tomuhle se vsechny chyby neslijou do jendoho univerzalniho `ClientException`, ale kazdy klient si zachova svuj exception scope (dokazes rozlisit, ze kteryho klienta exception vystrelila)

When the Keboola service clients were migrated onto the shared `keboola/php-api-client-base`, they all began throwing the single base `ClientException`, so a caller using several clients could no longer tell *which* client failed — previously each client shipped its own exception type and the class alone identified the source. This change restores per-client identification and, while we're here, gives the base exception structured HTTP context.

`ApiClient` gains a facade-supplied `exceptionClass` argument — a `class-string<ClientException>`, default `ClientException::class`. The base instantiates it **inline at each throw site**, so each service client can have the base throw its own `XClientException extends ClientException`: consumers `catch (VaultClientException)` for one client or `catch (ClientException)` for any Keboola client, and the stack trace is unchanged from before (inline `new`, no factory/helper frame — a subclass that wants richer construction does it in its own constructor, which receives `($message, $code, $previous, $statusCode, $responseBody)`). `ClientException` now also carries `getStatusCode()` / `getResponseBody()`, populated at every throw site (null for transport-level failures with no response). Everything is additive and back-compatible — the default path behaves exactly as before.

Key changes:
* `ApiClient`: new optional last ctor arg `exceptionClass` (`class-string<ClientException>`, default `ClientException::class`), instantiated **inline** at each throw site.
* `ClientException`: optional `statusCode`/`responseBody` (4th/5th ctor args) + `getStatusCode(): ?int` / `getResponseBody(): ?string`; `code` still equals the HTTP status for back-compat.
* Clean trace preserved by inline instantiation — no factory, closure, `makeException()` helper, or static named constructor (any of which would move the origin into themselves and add a trace frame). Customization is via the client's own subclass constructor.
* Per-client exception subclasses (e.g. `VaultClientException`) are intentionally **not** in this PR — they land in each client's own PR.

Design: `docs/superpowers/specs/2026-06-16-client-exception-model-design.md` (kept on the `pepa/common-api-lib-docs` branch, out of this PR).

None.

## Impact analysis
No end-user impact. Additive and fully back-compatible: existing callers keep receiving `ClientException`, and clients opt in by adding their own subclass and passing it. Base-lib minor release.

## Change type
Improvement

## Justification
Restore per-client exception identification that was lost when clients migrated onto the shared base, and expose HTTP status/body for programmatic error handling.

## Deployment
Merge & automatic deploy. Tag `php-api-client-base/1.1.0` to publish to Packagist.

## Rollback plan
Revert of this PR.

## Post release support plan
None.

### DIFF
--- a/libs/php-api-client-base/src/ApiClient.php
+++ b/libs/php-api-client-base/src/ApiClient.php
@@ -25,9 +25,17 @@ class ApiClient
     private readonly GuzzleClient $httpClient;
     private readonly ErrorMessageResolverInterface $errorMessageResolver;
 
+    /** @var class-string<ClientException> */
+    private readonly string $exceptionClass;
+
     /**
      * @param non-empty-string|null $baseUrl
      * @param list<int> $retryableStatusCodes Non-5xx status codes to also retry (e.g. [429]).
+     * @param class-string<ClientException> $exceptionClass Exception class thrown on failure; a
+     *     ClientException subclass lets callers identify the failing service. It is instantiated
+     *     inline at each throw site so the exception originates there (clean stack trace). A
+     *     subclass that wants richer construction does it in its own constructor, which receives
+     *     ($message, $code, $previous, $statusCode, $responseBody).
      */
     public function __construct(
         ?string $baseUrl,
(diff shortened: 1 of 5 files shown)


romantmb (approved):
👍

PR #511, opened 2026-06-08
feat(input-mapping): load workspace tables via input-mapping-load endpoint (AJDA-2841)

odinuv:
## Link to issue

https://linear.app/keboola/issue/AJDA-2841/input-mapping-refactor-before-removal-of-the-load-decider

This is the follow-up the AJDA-2841 refactor unblocked: the clone/copy/view decision is delegated to Connection and the client-side decider is removed.

## Description

Workspace table loading is switched from the SAPI `workspaces/{id}/load` / `load-clone` endpoints to the new `workspaces/{id}/input-mapping-load` endpoint, which accepts the `keboola/input-mapping` `Configuration\Table` payload (snake_case) unchanged and resolves the load type per item server-side (CLONE > VIEW > COPY, with feature gating by `bigquery-default-im-view` to switch VIEW > CLONE > COPY). The job echoes the resolved `loadType` per item back in `operationParams`.

Concretely:
- **Load-type decision removed from the library.** `AbstractWorkspaceStrategy::prepareAndExecuteTableLoads` now forwards each table's parsed configuration as-is (incl. any explicit `load_type`) to `queueInputMappingLoad`. The client-side decider and its supporting types are deleted: `LoadTypeDecider`, `WorkspaceLoadPlan`, `WorkspaceTableLoadInstruction`, `WorkspaceLoadType`, plus the `prepareTableLoadsToWorkspace`/`executeTableLoadsToWorkspace`/`decideTableLoadMethod` methods. The old `/load` and `/load-clone` calls are no longer used by the library.
- **Kept client-side** (the bits the endpoint can't do): resolving the adaptive `changed_since` marker to a concrete timestamp against the input state, and rejecting the `days` option on workspace loads.
- **`load_type` is case-insensitive** (normalized to upper-case in config parsing); the legacy `use_view: true` continues to normalize to `load_type: VIEW`.
- **`source_search` removed** — the `source_search` config node and the `TableDefinitionResolver` that resolved it to a concrete `source` are deleted; `source` is now required. (Already removed from all relevant configurations in https://linear.app/keboola/issue/AJDA-2879/zbavit-se-source-search-ktery-se-nepouziva-a-cely-to-desne-komplikuje.)
- CI: input-mapping is removed from the `ALLOW_FAIL` allow-list now that the suite is green again.

## Justification

See the linked issue. Delegating the load-type decision to Connection removes a duplicated, drift-prone client-side decider and lets the library pass its own config through unchanged.

## Plans for Customer Communication

None.

[...]

PR #523, opened 2026-06-17
[AJDA-2322] Fix affected-libraries detection after force-push/rebase

ondrajodas:
## Problem

When a branch is force-pushed (e.g. after a rebase), `github.event.before` points to the pre-rewrite tip, which becomes an orphaned commit not reachable from any ref. `actions/checkout` (even with `fetch-depth: 0`) only fetches ref-reachable history, so that SHA is absent from the runner's clone:

```
fatal: bad object 21b5f3b7...
affected-libraries failed, falling back to all: git diff failed (exit 128)
```

The `detect-changes` step then fell back to testing **all 22 libraries**, even when the branch only touched one (observed on PR #511, which changed only `libs/input-mapping/**`).

## Fix

`bin/ci/affected-libraries.php` now resolves the diff base more robustly:

- `github.event.before` stays the primary base — a normal push still diffs only its own commits (base-branch agnostic).
- When `before` is empty / all-zeros / **not present** in the checkout (force-push, rebase, first push), fall back to `merge-base origin/main HEAD` (tries `origin/main`, then `main`) and diff against that — so only the branch's real changes are resolved, not the whole monorepo.
- A `merge-base == HEAD` guard returns "no usable base" (→ test everything), which keeps pushes to `main` correct.
- Diffing against `main` only ever over-includes for branches stacked on another branch; it never under-tests.

The runner is wrapped in `resolveAffected()` behind a CLI guard so the logic is unit-testable.

## Tests

New `bin/ci/tests/AffectedLibrariesScriptTest.php` runs against real temporary git repos and covers: valid `before`, orphan/bad `before` (the reproduced bug), zero `before`, push to `main`, and infra changes. `composer ci` (phpcs + phpstan max + phpunit) passes.

### DIFF
--- a/bin/ci/affected-libraries.php
+++ b/bin/ci/affected-libraries.php
@@ -8,7 +8,6 @@ use Keboola\PlatformLibrariesCi\AffectedLibrariesResolver;
 require __DIR__ . '/vendor/autoload.php';
 
 $repoRoot = dirname(__DIR__, 2);
-$base = $argv[1] ?? '';
 
 /**
  * @return array<string, array{name: string, devDeps: list<string>}>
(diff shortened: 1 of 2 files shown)


ondrajodas:
```
# Release Notes
CI change only. The `detect-changes` step that selects which libraries to test/publish
no longer falls back to testing all 22 libraries after a force-push/rebase. It now
diffs against the merge-base with main and tests only the libraries the branch actually
changed. No runtime/library behavior changes.

None needed — internal CI tooling, no customer-facing impact.

# Impact Analysis
Affects only `bin/ci/affected-libraries.php` (the GitHub Actions detect-changes gate).
Worst case on any bug is over-selection (testing too many libs), which is safe and was
the previous behavior anyway; the change never under-selects (merge-base with main is a
superset of the branch changes). No production services, no library code touched.

# Change Type
Bug fix (CI tooling).

# Justification
After a force-push/rebase, github.event.before points to an orphaned commit absent from
the checkout, so `git diff before..HEAD` fails and CI tested every library (observed on
PR #511, which only changed libs/input-mapping). Linear: AJDA-2322.

# Deployment Plan
Merge to main. No config, secrets, or migrations. Takes effect on the next push-triggered
CI run.

# Rollback Plan
Revert this PR. The previous fallback-to-all behavior returns; no state to clean up.

# Post-Release Support Plan
Watch the first few feature-branch CI runs: confirm "Affected:" lists only the changed
libraries (not all 22) and that the merge-base fallback engages on force-pushed branches.
```

ondrajodas:
@MiroCillik můžeš prosím? 🙏🙏🙏🙏

odinuv (approved):
asi v poho

PR #524, opened 2026-06-22
PAT-1866 Migrate sync-actions-api-php-client onto php-api-client-base

pepamartinec:
## Release Notes
https://linear.app/keboola/issue/PAT-1866

Migrate `sync-actions-api-php-client` onto the shared `keboola/php-api-client-base`, matching the migrations already done for the vault, sandboxes, and git-service clients. The hand-rolled Guzzle client — its auth/retry/logging middleware stack and Symfony-validator constructor — is replaced by the base `ApiClient`, so transport, retries, error handling, and authentication are now shared across clients.

The public `Client` is renamed to `SyncActionsApiClient` with an unpacked constructor (individual params instead of an `ApiClientConfiguration` object), authenticates via the base `StorageApiTokenAuthenticator`, and keeps the historical default of 10 retries. Failures throw a per-client `SyncActionsClientException` (extends the base `ClientException`), formatted by a small `SyncActionsErrorMessageResolver`.

A sync action returns an arbitrary, component-defined payload that callers read as `$response->data->whatever`, so `ActionResponse.data` stays a deep `stdClass`. `callAction()` decodes the raw response body with `json_decode(..., false)` to preserve the object/array distinction — the base client's array decode would collapse `{}` to `[]` and renumber integer-keyed objects. That needs a small, backward-compatible base-lib change, included here since it's trivial: `ApiClient::sendRequest()` now returns the `ResponseInterface` instead of `void` (and the now-redundant private `doSendRequest()` is folded into it). `getActions()` still uses the array-based `sendRequestAndMapResponse` (its payload is a flat list of strings).

Key changes:
* `php-api-client-base`: `ApiClient::sendRequest()` returns the `ResponseInterface` (was `void`; `doSendRequest()` merged in) — backward-compatible.
* Rename `Client` → `SyncActionsApiClient`; delegate transport to the base `ApiClient`.
* Add `SyncActionsClientException` + `SyncActionsErrorMessageResolver`.
* `callAction()` decodes the raw body to a faithful `stdClass`; `ActionResponse` is a plain DTO.
* Drop `ApiClientConfiguration`, `StorageApiTokenAuthenticator`, `RetryDecider`, and the local `ResponseModelInterface`/`ClientException` (now provided by the base lib).
* Depend on `keboola/php-api-client-base`; drop `symfony/validator` and `symfony/config`.

None.

## Impact analysis
Breaking API change for consumers of `sync-actions-api-php-client` (notably `connection`): the exception type is renamed (`Keboola\SyncActionsClient\Exception\ClientException` → `SyncActionsClientException`), the constructor signature is unpacked (no `ApiClientConfiguration`), and the client class is renamed. Response reading (`$response->data->...`) is unchanged and now correct for empty/integer-keyed objects. The `php-api-client-base` change is backward-compatible (a `void` method now returns a value), so the other clients are unaffected. Requires a major `sync-actions-api-php-client` bump and a minor `php-api-client-base` bump.

## Change type
Refactoring

## Justification
Consolidate Keboola PHP API clients onto the shared `php-api-client-base` (PAT-1866).

## Deployment
Merge & automatic deploy.

## Rollback plan
Revert of this PR.

## Post release support plan
None.


janvanicek (approved):
lgtm

PR #526, opened 2026-06-22
PAT-1866 Return the response from ApiClient::sendRequest()

pepamartinec:
## Release Notes
https://linear.app/keboola/issue/PAT-1866

`ApiClient::sendRequest()` previously returned `void`, discarding the response. This changes it to return the `ResponseInterface` — a backward-compatible widening, so existing callers that ignore the return value are unaffected.

This lets a client decode the raw response body itself when the array-based `sendRequestAndMapResponse()` is unsuitable — e.g. payloads that must preserve the JSON object/array distinction, which an associative-array decode would lose (`{}` collapses to `[]`, integer-keyed objects get renumbered). The first consumer is the `sync-actions-api-php-client` migration.

Now that `sendRequest()` returns the response, the private `doSendRequest()` was a redundant indirection — it is folded into `sendRequest()`, which gains the optional per-request `$options` (already public via `sendRequestAndMapResponse()`).

None.

## Impact analysis
No end-user impact. Backward-compatible for all existing consumers — a `void` method now returns a value callers may ignore. The four already-migrated clients (vault, sandboxes, git-service, configuration-variables-resolver) are exercised by CI here and are unaffected.

## Change type
Improvement

## Justification
Enable clients to decode raw response bodies, needed by the sync-actions migration (PAT-1866).

## Deployment
Merge & automatic deploy.

## Rollback plan
Revert of this PR.

## Post release support plan
None.

### DIFF
--- a/libs/php-api-client-base/src/ApiClient.php
+++ b/libs/php-api-client-base/src/ApiClient.php
@@ -113,9 +113,22 @@ class ApiClient
         ]);
     }
 
-    public function sendRequest(RequestInterface $request): void
+    /**
+     * @param array<string, mixed> $options
+     */
+    public function sendRequest(RequestInterface $request, array $options = []): ResponseInterface
     {
-        $this->doSendRequest($request);
+        try {
+            return $this->httpClient->send($request, $options);
+        } catch (RequestException $e) {
+            throw $this->processRequestException($e);
+        } catch (GuzzleException $e) {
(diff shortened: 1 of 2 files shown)


pepamartinec:
Folded into #524 — the base-client change (`ApiClient::sendRequest()` returning the response + `doSendRequest()` merge) is trivial, so it ships in the single sync-actions PR rather than separately. Closing.

PR #528, opened 2026-06-22
fix(ci): key-pair workspace login for tests + fast-import-safe tag stripping in split-repo

zajca:
Combined CI/infra fix (folds in #529).

## 1. Test Snowflake workspaces use key-pair login

Snowflake no longer allows creating workspace users with the legacy service account type, which is what the empty/default workspace login type maps to. Every workspace-backed output-mapping functional test failed with:

> Snowflake no longer allows creating workspace users with the legacy service account type ("LEGACY_SERVICE").

`AbstractTestCase::initWorkspace()` now requests `snowflake-service-keypair` for Snowflake test workspaces and registers a generated public key (mirroring the production `WorkspaceProvider`). BigQuery keeps the default login type.

## 2. split-repo.sh: avoid fast-import ref collision when stripping foreign tags

The `publish` job failed for every library with:

```
error: multiple updates for ref 'refs/tags/SKIP' not allowed
```

`split-repo.sh` renamed every tag outside the library's prefix to a single `refs/tags/SKIP` before deleting it. `git fast-import` rejects multiple updates to the same ref, so once two foreign tags are annotated (`k8s-client/4.7.0`, `k8s-client/4.7.1`) the split aborts. Foreign tags are now renamed to a unique throwaway ref (`refs/tags/__split_skip__/<orig>`) and bulk-deleted.

Verified against a fresh mirror of the live monorepo (filter-repo via apt as in CI): old callback fails, new one succeeds and leaves only the prefix-stripped library tags.

## Note

This PR touches `bin/**`, so CI runs all 22 libraries. Other Snowflake-workspace libraries (e.g. staging-provider, query-service-api-client) may still fail on the same legacy-login issue until their test helpers get the same key-pair fix.

### DIFF
--- a/libs/output-mapping/tests/Needs/TestSatisfyer.php
+++ b/libs/output-mapping/tests/Needs/TestSatisfyer.php
@@ -57,11 +57,11 @@ class TestSatisfyer
     ): string {
         $bucketId = self::getBucketIdByDisplayName($clientWrapper, $bucketName, $stage);
         if ($bucketId !== null) {
-            $tables = $clientWrapper->getTableAndFileStorageClient()->listTables($bucketId, ['include' => '']);
-            foreach ($tables as $table) {
-                $clientWrapper->getTableAndFileStorageClient()->dropTable($table['id']);
-            }
-            return $bucketId;
+            // Force-drop and recreate the bucket instead of emptying it table by table. Dropping the bucket
+            // drops the whole backend schema, which also removes backend objects orphaned by an interrupted
+            // run (e.g. a build cancelled mid-test): those are not tracked as Storage tables, so emptying
+            // would leave them behind and the next table creation would fail with "object already exists".
+            $clientWrapper->getTableAndFileStorageClient()->dropBucket($bucketId, ['force' => true]);
         }
         return $clientWrapper->getTableAndFileStorageClient()->createBucket(
             name: $bucketName,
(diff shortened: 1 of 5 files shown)


libs/output-mapping/tests/Needs/TestSatisfyer.php
```diff
@@ -57,11 +57,11 @@ private static function ensureEmptyBucket(
     ): string {
         $bucketId = self::getBucketIdByDisplayName($clientWrapper, $bucketName, $stage);
         if ($bucketId !== null) {
-            $tables = $clientWrapper->getTableAndFileStorageClient()->listTables($bucketId, ['include' => '']);
-            foreach ($tables as $table) {
-                $clientWrapper->getTableAndFileStorageClient()->dropTable($table['id']);
-            }
-            return $bucketId;
+            // Force-drop and recreate the bucket instead of emptying it table by table. Dropping the bucket
```
odinuv, line 60: No klidně, ale tohle tak nebylo pro srandu králíkům, ale protože na connection se rozbíjely bucket permissions, když se bucket smazal a hned založil, tak uvidíme :)

odinuv (approved):

libs/output-mapping/tests/Needs/TestSatisfyer.php
zajca: tak uvidíme jestli to rozbijeme :D

PR #531, opened 2026-06-23
PAT-1895 Build Storage client from resolved StorageApiToken

pepamartinec:
## Release Notes
https://linear.app/keboola/issue/PAT-1895/storage-api-php-client-branch-wrapper-do-not-use-storage-token-from

> [!NOTE]
> Implementace `StorageClientApiFactory` + controller argument resolver.
> * factory se bude pouzivat misto existujici `StorageClientRequestFactory`, ktera nefunguje s PAT tokenama
> * resolver usnadnuje pouziti pro nejcastejsi use-case - chci Storage API clienta pro aktualni Storage token

Adds `Keboola\ApiBundle\StorageApiClient\StorageClientApiFactory`, a Storage API client factory that builds a `ClientWrapper` from an already-resolved `StorageApiToken` object instead of re-reading the token from the request headers. The branch-wrapper's `StorageClientRequestFactory` derives the token straight from `Authorization: Bearer` / `X-StorageApi-Token`, which breaks for Connection programmatic tokens (`kbc_pat_*` / `kbc_at_*`): the header carries the programmatic token, while the real legacy Storage token only exists on the `StorageApiToken` that api-bundle's authenticator resolved via the Manage API exchange. api-bundle is the right home for this factory — it already resolves the `StorageApiToken` and already depends on `keboola/storage-api-php-client-branch-wrapper`.

The factory takes the token value from `StorageApiToken::getTokenValue()`, pins `authType` to `AuthType::STORAGE_TOKEN` (the resolved value is always a legacy Storage token), and resolves the run id from `X-KBC-RunId` with the usual generator/`uniqid` fallback. It is intentionally independent of `StorageClientRequestFactory` — it owns its own `ClientOptions` and neither factory references the other. This PR ships only the class and its unit tests; DI auto-registration in `KeboolaApiExtension` and the consumer-facing usage docs are a deliberate follow-up step, so nothing in the container consumes the factory yet.

None.

## Impact analysis
No end-user impact. The factory is not registered in the container yet (no DI wiring), so no code path consumes it; this PR only adds a new class and its tests.

## Change type
New feature

## Justification
Multi-project / PAT support: downstream services must build Storage clients from the resolved `StorageApiToken`, not from request headers that carry a programmatic token.

## Deployment
Merge & automatic deploy.

## Rollback plan
Revert of this PR.

## Post release support plan
None.


libs/api-bundle/src/StorageApiClient/StorageClientApiFactory.php
```diff
@@ -0,0 +1,54 @@
...
+
+namespace Keboola\ApiBundle\StorageApiClient;
+
+use Keboola\StorageApiBranch\ClientWrapper;
+use Keboola\StorageApiBranch\Factory\AuthType;
+use Keboola\StorageApiBranch\Factory\ClientOptions;
+use Keboola\StorageApiBranch\StorageApiToken;
+use Symfony\Component\HttpFoundation\Request;
+
+class StorageClientApiFactory
```
pepamartinec, line 13: Naming is hard. Nasleduju pattern existujicich factory (`StorageClientRequestFactory`, `StorageClientPlainFactory`)

romantmb (approved):
LGTM

PR #532, opened 2026-06-24
test(output-mapping,query-service-api-client,staging-provider): AJDA-2820 use loginType none for test workspaces

odinuv:
## Link to issue

https://linear.app/keboola/issue/AJDA-2820/update-job-runner

## Description
Trošku jsem poleštil tady to https://github.com/keboola/platform-libraries/pull/528/changes#diff-97d65c973832a3c296970b648078c6c8251b01e8c35ee84af2d611a0bc680aa9R158

[#528](https://github.com/keboola/platform-libraries/pull/528) reworked the Snowflake test-workspace setup because Snowflake no longer allows creating workspace users with the legacy service-account type (the empty/default `loginType`). It worked around this by requesting a `SNOWFLAKE_SERVICE_KEYPAIR` login and registering a generated public key.

That heavier approach isn't needed in these three test setups — none of them ever connects to the workspace with its own credentials:

- **output-mapping** `AbstractTestCase` — `workspaceCredentials` is never read; data is loaded/cloned server-side via the Storage API by workspace ID.
- **query-service** `BaseFunctionalTestCase` — SQL runs through the Query Service broker (`submitQueryJob`); the test never opens a workspace connection.
- **staging-provider** `WorkspaceProviderFunctionalTest::createWorkspace` (default branch) — used only by tests that inspect workspace metadata or exercise cleanup. The explicit key-pair credential-reset test sets `loginType` itself and is unaffected.

So all three now create the workspace with `WorkspaceLoginType::NONE` (a schema with no Snowflake login), matching the pattern already in use by `input-mapping`, and the now-unnecessary key-pair generation is dropped.

## Justification

Simplifies the merged #528 test setup and aligns it with `input-mapping`. See the linked Linear issue.

None.

## Impact Analysis

Test-only changes. The `query-service` case is the only behavioral assumption worth watching: it relies on the Query Service being able to execute SQL against a `none`-login workspace via its own privileged connection. Confirmed green by the functional CI suite for these libraries.

## Deployment Plan

Release a tagged version (libraries).

## Rollback Plan

Revert this PR.

## Post-Release Support Plan

None.

### DIFF
--- a/libs/output-mapping/tests/AbstractTestCase.php
+++ b/libs/output-mapping/tests/AbstractTestCase.php
@@ -5,7 +5,6 @@ declare(strict_types=1);
 namespace Keboola\OutputMapping\Tests;
 
 use InvalidArgumentException;
-use Keboola\KeyGenerator\PemKeyCertificateGenerator;
 use Keboola\OutputMapping\Staging\StrategyFactory;
 use Keboola\OutputMapping\TableLoader;
 use Keboola\OutputMapping\Tests\Needs\TestSatisfyer;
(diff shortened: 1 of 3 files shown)


ondrajodas (approved):
https://www.youtube.com/watch?v=JQlXYoxEBGM

PR #539, opened 2026-07-09
PAT-1839 Make api-bundle Storage client options configurable

pepamartinec:
## Release Notes
https://linear.app/keboola/issue/PAT-1839/support-pat-tokens-in-queue-job-queue

`api-bundle` builds the base `ClientOptions` for `StorageClientApiFactory` internally, wiring only the Connection URL, the shared `@logger` and the `app_name` user agent — every other option (notably `backoffMaxTries`) was left at its default and could not be adjusted by consumers without patching the bundle. This adds an optional `keboola_api.storage_client_options` node so services can tune the base options from their own configuration.

The node accepts either a **service id** (string) referencing a `ClientOptions` service, or an **object** of individual YAML-expressible options: `backoff_max_tries`, `aws_retries`, `aws_debug`, `retry_on_maintenance`, `use_branch_storage`, `user_agent`, and `logger` (a service id for a PSR `LoggerInterface`). In both forms the values are *merged onto* the bundle-built base via `ClientOptions::addValuesFrom()`, so consumers never have to reproduce the `ServiceClient` URL resolution. The two forms are mutually exclusive, and closures (`jobPollRetryDelay` / `runIdGenerator`) or `BackendConfiguration` are reachable only through the service form. When the node is omitted the base options are wired exactly as before.

None.

## Impact analysis
No end-user impact. The `storage_client_options` node is optional; when omitted the base `ClientOptions` for `StorageClientApiFactory` are built byte-for-byte as before.

## Change type
New feature

## Justification
Lets services adopting api-bundle for PAT-token support (e.g. `job-queue`) tune the Storage client — `backoffMaxTries` in particular — without patching the bundle.

## Deployment
Merge & automatic deploy.

## Rollback plan
Revert of this PR.

## Post release support plan
None.


libs/api-bundle/src/DependencyInjection/Configuration.php
```diff
@@ -25,6 +25,42 @@ public function getConfigTreeBuilder(): TreeBuilder
...
+                        ->ifString()
+                        ->then(fn(string $v) => ['service' => $v])
+                    ->end()
+                    ->children()
+                        ->scalarNode('service')
+                            ->cannotBeEmpty()
+                            ->info('Service id of a ClientOptions instance merged onto the base options.')
+                        ->end()
+                        ->integerNode('backoff_max_tries')->end()
+                        ->integerNode('aws_retries')->end()
```
janvanicek: Optional: add `->min(0)` to these integer nodes to reject nonsensical negative values, e.g. `->integerNode('backoff_max_tries')->min(0)->end()`.

janvanicek (approved):
lgtm

PR #541, opened 2026-07-15
chore(sandboxes-service-api-client): drop provisioningStrategy field

pepamartinec:
## Release Notes
https://github.com/keboola/sandboxes-service/pull/631

The sandboxes-service is removing the `provisioningStrategy` field from its Apps API response. This PR drops the field from the `sandboxes-service-api-client` so the client no longer reads or exposes it. Without this change `App::fromResponseData()` would fail with an undefined-array-key error the moment the API stops returning the field.

The `provisioningStrategy` property, its constants (`PROVISIONING_STRATEGY_JOB_QUEUE`, `PROVISIONING_STRATEGY_OPERATOR`, `VALID_PROVISIONING_STRATEGIES`), the required-property entry, the getter/setter, and the `toArray()` key are all removed, together with the corresponding test fixtures and assertions.

None.

## Impact analysis
The client no longer reads or exposes `provisioningStrategy`. This is required to stay compatible with the sandboxes-service API dropping the field; any consumer that read `getProvisioningStrategy()` must stop doing so.

## Change type
Maintenance

## Justification
Keep the API client compatible with the sandboxes-service API that is dropping the `provisioningStrategy` field.

## Deployment
Merge & automatic deploy.

## Rollback plan
Revert of this PR.

## Post release support plan
None.

### DIFF
--- a/libs/sandboxes-service-api-client/src/Apps/App.php
+++ b/libs/sandboxes-service-api-client/src/Apps/App.php
@@ -17,7 +17,6 @@ final class App implements ResponseModelInterface
         'configVersion',
         'state',
         'desiredState',
-        'provisioningStrategy',
     ];
 
     public const STATE_CREATED = 'created';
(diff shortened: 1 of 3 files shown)


janvanicek (approved):
LGTM

PR #542, opened 2026-07-17
PAT-1938 Switch Query API PHP client to the new base client

pepamartinec:
## Release Notes
https://linear.app/keboola/issue/PAT-1938/switch-query-api-php-client-to-the-new-base-client

Migrates `keboola/query-api-php-client` (`libs/query-service-api-client`) off its hand-rolled Guzzle 7 transport onto the shared `keboola/php-api-client-base`, matching the house pattern already used by `git-service-api-client`, `sync-actions-api-php-client` and the migrated `encryption-api-php-client`. This is the enabling step for PAT-token support across services — PATs are Storage API tokens and flow through the existing `X-StorageApi-Token` header, so no PAT-specific code is needed here. Because the constructor shape and the exception namespace change, this is a **major version bump**.

The `Client` now composes a private `Keboola\ApiClientBase\ApiClient`: `StorageApiTokenAuthenticator` for `X-StorageApi-Token`, a `QueryApiErrorMessageResolver` that preserves the current error contract (server `exception` field → message, HTTP status → code), and a house-style `Keboola\QueryApi\Exception\ClientException` that now also exposes `getStatusCode()` / `getResponseBody()`. The five response DTOs implement the base `ResponseModelInterface::fromResponseData`; the bespoke `RetryDecider`, `ResponseParser` and the `symfony/validator` dependency are removed. The full public surface — the five endpoint methods, `PaginationHelper`, `ResultHelper`, and every DTO getter/constructor — is preserved.

Key changes:
* `Client` constructor moves from `(array $config, array $options)` to flat-scalar `(string $baseUrl, string $storageToken, ?string $runId = null, ?LoggerInterface $logger = null, int $backoffMaxTries = 3, int $connectTimeout = 10, int $requestTimeout = 120, string $userAgent = ..., null|Closure|HandlerStack $requestHandler = null)`.
* Exception FQN moves from `Keboola\QueryApi\ClientException` to `Keboola\QueryApi\Exception\ClientException` (subclass of the base exception); the old `Keboola\QueryApi\Exception` base class is removed.
* Retry / timeouts / user-agent / logging now come from the shared base client (retry semantics unchanged: 5xx + transport errors, never 4xx, default 3 tries; the `X-KBC-RunId` header is preserved).
* `ext-json` is declared on `keboola/php-api-client-base` (owner of the `Json` helper) rather than duplicated on each consumer.
* Unit tests are decoupled from live credentials (the test bootstrap no longer hard-requires Storage env vars); functional tests are unchanged and still require CI-provided Storage credentials.

The branch also carries the design spec and implementation plan under `docs/superpowers/` — drop those commits if you'd rather not keep them in-tree.

None.

## Impact analysis
No end-user impact. This is an internal PHP client for the Query Service; no other library in the monorepo imports `Keboola\QueryApi\*`, so the breaking constructor and exception-namespace changes only affect external consumers, which pick them up with the major version bump.

## Change type
Refactoring

## Justification
PAT token support across services — switch the Query API client to the shared `keboola/php-api-client-base`.

## Deployment
Merge & automatic deploy. The new major version is published by pushing a `query-service-api-client/<major>.0.0` tag afterwards.

## Rollback plan
Revert of this PR.

## Post release support plan
None.


janvanicek (approved):
lgtm

PR #544, opened 2026-07-20
PAT-1896 Drop StorageClientRequestFactory dependency from api-bundle

pepamartinec:
## Release Notes
https://linear.app/keboola/issue/PAT-1896

`api-bundle` no longer depends on `Keboola\StorageApiBranch\Factory\StorageClientRequestFactory` from `keboola/storage-api-php-client-branch-wrapper`. That class was referenced as a service the bundle never registered, so every consuming app had to wire it up (with its own base `ClientOptions`) for `#[StorageApiTokenAuth]` to work — an undocumented footgun that also let token *verification* and the controller-facing Storage client drift apart on their base options. The package itself stays required: `ClientWrapper`, `ClientOptions`, `AuthType` and the base `StorageApiToken` are still used.

Verification now builds its Storage client from the bundle's own base `ClientOptions` via a new bundle-owned `RequestStorageClientFactory`, so verification and `StorageClientApiFactory` share one base-options service. `StorageApiTokenAuthenticator` classifies the incoming token exactly once — `extractCredential()` returns a `RequestTokenType` of `StorageToken` / `OAuthToken` / `Programmatic` — and dispatches each type to one dedicated `StorageApiTokenFactory` method. `ClientWrapper` construction, including run-id resolution, lives once in `StorageClientWrapperFactory`.

Key changes:
* Remove the `StorageClientRequestFactory` dependency and the implicit "consuming app must register it" contract.
* Add `RequestStorageClientFactory` (mockable verification seam), `StorageClientWrapperFactory` (single `ClientWrapper` builder), and the `RequestToken` / `RequestTokenType` classification the authenticator emits.
* Classify the request token once and map each type to one factory method: `createFromStorageToken()` / `createFromOAuthToken()` (both delegating to a shared private `verify()`) and `createFromProgrammaticToken()` — replacing the previous asymmetric `createFromRequest()` / `createFromProgrammaticToken()` split.
* Verification and the controller-facing client share one base `ClientOptions` service, so `storage_client_options` now applies to both.

None.

## Impact analysis
No end-user impact for the supported auth paths: `X-StorageApi-Token` and `Authorization: Bearer` behave identically (auth-type mapping and programmatic-token exchange are unchanged, and locked by existing tests). Two edge deltas on unsupported/malformed input, both still returning `401`: a non-`Bearer` `Authorization` value with no `X-StorageApi-Token` now reaches a Storage verify instead of an immediate local 401, and in functional tests the resolved auth type follows the request headers (as production already did) instead of a mock-baked value — the `AuthenticatorTestTrait` helpers require the request under test to carry the matching scheme.

## Change type
Refactoring

## Justification
Remove an implicit cross-package coupling (the unregistered `StorageClientRequestFactory` service) and de-duplicate token extraction and Storage client construction in api-bundle.

## Deployment
Merge & automatic deploy.

## Rollback plan
Revert of this PR.

## Post release support plan
None.

### DIFF
--- a/libs/api-bundle/README.md
+++ b/libs/api-bundle/README.md
@@ -149,9 +149,10 @@ which the bundle requires directly, so no extra installation is needed.
 
 When `#[StorageApiTokenAuth]` is enabled, type-hint
 `Keboola\ApiBundle\StorageApiClient\StorageClientApiFactory` on a controller argument; the bundle
-injects a factory already bound to the current request and the resolved `StorageApiToken`. Unlike the
-header-based `StorageClientRequestFactory`, it uses the token resolved by the authenticator, so it
-works for programmatic (`kbc_pat_*` / `kbc_at_*`) tokens too.
+injects a factory already bound to the current request and the resolved `StorageApiToken`. It uses
+the token resolved by the authenticator, so it works for programmatic (`kbc_pat_*` / `kbc_at_*`)
+tokens too, and shares the same base `ClientOptions` the bundle uses internally to verify the
+incoming token — no external Storage client factory needs to be registered by the consuming app.
 
 The base options are preconfigured by the bundle: the Connection (Storage API) URL from the
 `ServiceClient`, the shared logger, and the configured `app_name` as user agent. The run id comes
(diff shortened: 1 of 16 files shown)


pepamartinec:
Folded into #546, which now contains this change plus the auth-internals simplification as a single PR against main.

PR #547, opened 2026-07-23
PAT-1959 Own the runId generator; drop client-wrapper StorageClientRequestFactory

pepamartinec:
## Release Notes
https://linear.app/keboola/issue/PAT-1959/api-bundle-own-the-runid-generator-drop-client-wrapper

Related `storage-api-php-client-branch-wrapper` 7.0: https://github.com/keboola/storage-api-php-client-branch-wrapper/pull/39 (PAT-1954)

branch-wrapper 7.0 removed `ClientOptions::runIdGenerator` (and its `get/setRunIdGenerator()`) together with `Keboola\StorageApiBranch\Factory\StorageClientRequestFactory`, so api-bundle's `StorageClientRequestFactory::resolveRunId()` — which read the generator off `ClientOptions` — no longer compiles against it. The bundle now **owns** the generator: `StorageClientRequestFactory` takes it as an optional `?Closure(ClientOptions): string` constructor dependency and resolves from it, with the run id order unchanged — `X-KBC-RunId` header → configured generator → `uniqid('run-')`.

Because the generator is a closure and therefore not YAML-expressible, consumers inject it through a new `keboola_api.storage_client_options.run_id_generator: <service-id>` knob, wired in `KeboolaApiExtension` to the factory's `$runIdGenerator` argument. The referenced service must resolve to a `Closure` (exposed via a factory, since `Closure` is not directly constructible). It is grouped under `storage_client_options` for config cohesion but — unlike the other keys — is passed to the factory rather than merged into `ClientOptions`, so it is exempt from the `service`-vs-individual-options exclusivity and may accompany either form. This replaces the previous approach of setting `runIdGenerator` on a custom `ClientOptions` service, which the client wrapper no longer supports.

Key changes:
* `StorageClientRequestFactory` owns the run id generator (`?Closure` ctor dep); no more `ClientOptions::getRunIdGenerator()`.
* New `keboola_api.storage_client_options.run_id_generator` config knob + `KeboolaApiExtension` wiring.
* `keboola/storage-api-php-client-branch-wrapper` bumped to `^7.0`.
* README: dropped `runIdGenerator` from the custom-`ClientOptions`-service docs; added a "Customizing the run id" section.
* Audited all `StorageApiToken` / `ClientOptions`-with-token construction — every site already passes `authType`/`tokenType`, so 7.0's "token requires authType" enforcement is satisfied; no code changes were needed there.

None.

## Impact analysis
To be released as api-bundle **6.1.0**. It moves the `keboola/storage-api-php-client-branch-wrapper` requirement to `^7.0`; the only service affected by the runId-generator change is `runner-sync-api` (the sole user of a custom generator), migrated in lockstep via PAT-1960. Any service that injected a run id generator through a custom `ClientOptions` service must move to the `keboola_api.storage_client_options.run_id_generator` knob. No end-user (runtime) impact — run id resolution behaviour is unchanged.

## Change type
Refactoring

## Justification
Adapt api-bundle to storage-api-php-client-branch-wrapper 7.0, which removed `ClientOptions::runIdGenerator` and `Factory\StorageClientRequestFactory`.

## Deployment
Merge & automatic deploy.

## Rollback plan
Revert of this PR.

## Post release support plan
None.

### DIFF
--- a/libs/api-bundle/tests/DependencyInjection/KeboolaApiExtensionTest.php
+++ b/libs/api-bundle/tests/DependencyInjection/KeboolaApiExtensionTest.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Keboola\ApiBundle\Tests\DependencyInjection;
 
+use Closure;
 use Keboola\ApiBundle\DependencyInjection\KeboolaApiExtension;
 use Keboola\ApiBundle\Security\ApplicationToken\ManageApiClientFactory;
 use Keboola\ApiBundle\Security\StorageApiToken\StorageApiTokenAuthenticator;
@@ -12,11 +13,13 @@ use Keboola\ApiBundle\StorageApiClient\StorageClientApiFactoryResolver;
 use Keboola\ApiBundle\StorageApiClient\StorageClientRequestFactory;
 use Keboola\ManageApi\Client as ManageApiClient;
 use Keboola\ServiceClient\ServiceClient;
+use Keboola\StorageApiBranch\Factory\AuthType;
 use Keboola\StorageApiBranch\Factory\ClientOptions;
 use PHPUnit\Framework\TestCase;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Definition;
 use Symfony\Component\DependencyInjection\Reference;
+use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorageInterface;
 
 class KeboolaApiExtensionTest extends TestCase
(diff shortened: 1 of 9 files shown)


libs/api-bundle/tests/StorageApiClient/StorageClientRequestFactoryTest.php
```diff
@@ -52,9 +52,10 @@ public function testRunIdFallsBackToGeneratedValueWhenHeaderMissing(): void
 
     public function testRunIdGeneratorUsedWhenHeaderMissing(): void
```
janvanicek, line 53: Přidal bych test na ověření, že `run id generator must not be called when the X-KBC-RunId header is present`

janvanicek (changes requested):
Jen ten jeden test case, jinak ok

libs/api-bundle/tests/StorageApiClient/StorageClientRequestFactoryTest.php
pepamartinec: Added `testRunIdGeneratorNotCalledWhenHeaderPresent`: the injected generator calls `self::fail()` if invoked and the request carries `X-KBC-RunId: 42`, so it asserts the header wins and the generator is never called. Verified it fails with that message when the header short-circuit is removed. 👍

janvanicek (approved):

libs/api-bundle/tests/DependencyInjection/KeboolaApiExtensionTest.php
```diff
@@ -260,4 +263,102 @@ public function testStorageClientOptionsAbsentLeavesBaseUntouched(): void
             array_keys($base->getArguments()),
         );
     }
+
+    // -------------------------------------------------------------------------
+    // run_id_generator
+    // -------------------------------------------------------------------------
```
janvanicek, line 269: Ještě tohle nevim nevim teda

PR #550, opened 2026-07-27
feat(output-mapping): AJDA-2946 store description only when table flag is system-managed

zajca:
Resolves [AJDA-2946](https://linear.app/keboola/issue/AJDA-2946/output-mapping-store-description-only-when-table-flag-is-system).

## Why

Until now output mapping stored the description as `KBC.description` metadata and had no way to tell whether it may touch the description at all — writing it on every run would silently overwrite a description a user had set in the UI.

[DMD-1662](https://linear.app/keboola/issue/DMD-1662) (connection [#7754](https://github.com/keboola/connection/pull/7754)) added the table-level `isDescriptionSystemManaged` flag exactly for this decision. This PR makes output mapping act on it and store the description in the **native first-class `description` field**, which is what Input Mapping now reads primarily ([AJDA-2714](https://linear.app/keboola/issue/AJDA-2714)).

This supersedes #525, which solved the same problem by guessing from the presence of a `user` provider `KBC.description` row. The flag replaces the heuristic.

## Behaviour

| Situation | Where | What happens |
|---|---|---|
| Table is created by this run through a table definition | `TableDefinitionInterface::getRequestData()` | the description is part of the create payload — a fresh table is system-managed by the Storage default |
| Table is created by the load job itself (`CreateAndLoadTableTask`) | `LoadTableQueue::waitForAll()`, after a successful load | the description is stored once the table exists — there is no create payload to embed it into |
| Table already existed | `StoragePreparer`, before the load | the description is stored **only** when `isDescriptionSystemManaged` is `true`; otherwise it is discarded and an `info` message is logged |

Every path that touches an existing table goes through one flag-gated entry point (`TableDescriptionModifier::updateDescriptions()`) taking the table detail, so the guard cannot be bypassed. A table created by the run is system-managed by definition and has nothing to diff against.

Only real changes are sent: the current descriptions are read from the table definition and compared with what the configuration produced. When nothing changed, no API call is made at all — so a steady-state run costs **zero** extra calls.

Descriptions of a table created through a table definition are **embedded in the create payload**, so no extra blocking table-definition update job runs after the load. This covers all three create variants — typed from `schema`, typed from legacy column metadata, and non-typed from manifest `columns`. The create endpoint nests the column description in `columns[].definition.description`; a `definition` holding only a description carries no type, so the table stays non-typed.

`FailedLoadTableDecider::decideTableDelete()` therefore ignores the `KBC.description` metadata row written under the **`storage`** provider. A description in the create payload makes Storage write that row at create time, before any data is loaded, so it says nothing about the table having been used and must not block the drop of a failed empty table. A failed load also drops its pending description, so the "was not stored" warning stays reserved for descriptions lost by mistake.

## `KBC.description` is no longer written as component metadata

Writing the native description makes Storage mirror it into a `KBC.description` metadata row under the **`storage`** provider — for the table and for every column (`TableDefinitionUpdateService::persistTableDescription()` and `StorageColumnDefinitionSynchronizer::syncColumnDescriptions()`). Output mapping's own componentId row was therefore a duplicate, and on a **user-managed** table it published a competing description while the user's value sat under `storage` — for anyone reading metadata, output mapping had overwritten the user's description. That failed the acceptance criterion, so the key is no longer written as component metadata. Resolution and stripping live in `Writer/Helper/DescriptionHelper`.

A column whose only metadata was the description is still reported by `getColumnMetadata()` with an empty metadata list, because the set of columns to add to an existing table is derived from those keys as well (`TableStructureModifier`); `ColumnsMetadata` now skips such empty entries instead of sending an empty metadata write.

**BC note:** consumers that read `KBC.description` and filter on `provider == componentId` will stop seeing it. The row under the `storage` provider remains, so consumers matching on the key are unaffected — including the input mapping manifest, which does not filter by provider.

## Notes for review

- `TableInfo::isDescriptionSystemManaged()` defaults to `true` when the field is missing from the table detail, matching the platform default (and the DMD-1662 migration, where existing rows read as system-managed).
- An empty description is treated as no description, so it never clears a value stored in Storage.
- `table_metadata` and the schema column `metadata` are `variableNode`s, so a non-object value can reach the code. It is a configuration error and is reported as `InvalidOutputException` rather than dropped.
- Column descriptions are filtered against the columns the table really has; skipped ones are logged as a warning rather than failing the request (Storage answers 404 for an unknown column and would reject the whole patch, table description included).
- A `ClientException` outside the 4xx range is propagated instead of being wrapped, so a Storage outage and a connection error (which carries no HTTP status) both stay a retryable application error — consistent with the metadata path in `LoadTableQueue`.
- **Storing a column description on a typed table resyncs its datatype metadata.** `TableDefinitionUpdateService::updateTableDefinitionOnStorageBackend()` takes the `syncTypedColumnInfo()` branch for any `columns[]` patch on a typed table — the description-only shortcut is gated on `!$table->isTypedTable()` — which re-reads every column from the backend and replaces `KBC.datatype.type|nullable|basetype|length|default` under the `storage` provider. That is a no-op wherever the recorded metadata already matches the physical column, so it only shows up where the two had drifted: a column added by basetype records the requested Snowflake alias (`TIMESTAMP`) while the column really is `TIMESTAMP_LTZ` with length `9`. `testWriterUpdateTableDefinitionWithBaseTypes` asserts the resynced values. A table-level description alone never triggers this (`$columns === []` returns early), and a non-typed table takes the shortcut. The drift itself comes from `Storage_Job_TableColumnAdd::addTableColumnToTypedTable()`, which writes back the requested definition instead of re-reading the backend — worth a separate look by the Storage team.
- `keboola/storage-api-client` bumped `^18.5` → `^18.9` for `updateTableDefinition()`.
- `LoadTableTaskCreator::create()` returns a `LoadTableTaskResult` carrying the descriptions that could not be embedded; `LoadTableQueue` got a new optional 4th constructor argument, so it stays backwards compatible.
- The new functional test is excluded from the `main-writer-tests-2` suite so it does not run in two CI jobs in parallel against the same project (bucket name collision).

## Testing

- phpcs + phpstan (level max) clean; unit tests green.
- Descriptions are asserted inside the tests which already exercise each path, so the coverage adds no extra Storage jobs: typed create and update in `TableDefinitionTest`, a column added by the same run in `TableDefinitionV2Test::testAddMissingColumnTableDefinition`, BigQuery in `TableDefinitionV2BigQueryTest::testWriteTableOutputMappingExistingTable`, and workspace unload in `WriterWorkspaceTest::testSnowflakeTableOutputMapping`.
- `tests/Writer/TableDescriptionWriterTest.php` keeps what needs real Storage: a created table gets its description, a system-managed one is updated on the next run, a user-managed one is left alone, a table created by the load job behaves like the others, a description dropped from the configuration does not clear the stored value, a description of a column the data does not have is reported and skipped, and a failed load still drops the freshly created table.
- The repeated-unchanged-run case is what caught the most serious defect of the first revision: Storage answers 400 `No table definition changes were provided.` for a patch with no effective change, so every second run would have killed output mapping — for an existing table before any data was loaded.


libs/output-mapping/src/Mapping/MappingFromProcessedConfiguration.php
```diff
@@ -168,26 +177,104 @@ public function getDistributionKey(): array
...
+                }
+            }
+
+            return $descriptions;
+        }
+
+        // read from the raw configuration - getColumnMetadata() has the description stripped out
+        $columnMetadataFromConfiguration = $this->mapping['column_metadata'] ?
+            RestrictedColumnsHelper::removeRestrictedColumnsFromColumnMetadata($this->mapping['column_metadata']) :
+            [];
```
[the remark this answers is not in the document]
zajca: The caution is right in general, but it applies to `table_metadata`, not to `column_metadata`.

`$this->mapping` is not the raw configuration — the constructor assigns it as `(new Configuration())->parse([$mapping])` (`MappingFromProcessedConfiguration.php:27`), and every call site goes through that constructor, so an unparsed array cannot reach this method.

`column_metadata` is declared as a prototyped array node (`Configuration/Table/BaseConfiguration.php:177-188`), which Symfony builds as a `PrototypedArrayNode`. That class hard-codes `hasDefaultValue()` to `true` (`symfony/config/Definition/PrototypedArrayNode.php:101-104`) with a `[]` default, so `ArrayNode::finalizeValue()` always writes the key into the finalized result even when the user did not supply it. Parsing a mapping with only `destination` set confirms it:

```
array_key_exists('column_metadata', $parsed)  => true   ([])
array_key_exists('table_metadata',  $parsed)  => false
```

It also cannot be non-array — an `ArrayNode` throws `InvalidTypeException` on a non-array value. The node that really is variable and really can be absent is `table_metadata` (`BaseConfiguration.php:194`, a `variableNode`), and that one is already guarded with `?? []` plus an `is_array()` check in `getTableDescription()` (`MappingFromProcessedConfiguration.php:220-222`).

Finally, this ternary is not new logic — it is the same expression `getColumnMetadata()` has used on `main` for a long time (`MappingFromProcessedConfiguration.php:142-144`); it is not part of this diff.

No change made.

libs/output-mapping/src/DeferredTasks/FailedLoadTableDecider.php
```diff
@@ -29,6 +30,16 @@ public static function decideTableDelete(
...
 
+        // A description passed in the create-table-definition payload makes Storage write a KBC.description
+        // row under the "storage" provider at create time, i.e. before anything is loaded. It therefore says
+        // nothing about the table having been used and must not stop the drop of a failed empty table. This
+        // is intentionally outside the isTyped branch above - the non-typed create-table-definition payload
+        // (TableDefinitionFromColumns) can carry a description too.
+        $metadata = array_filter(
+            $metadata,
+            fn($m) => $m['key'] !== DescriptionHelper::DESCRIPTION_METADATA_KEY || $m['provider'] !== 'storage',
+        );
```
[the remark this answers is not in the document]
zajca: The filter introduces no new failure surface, and the scoping the comment asks for is already what it does.

On the shape assumption: `$metadata = $tableInfo['metadata']` and the `KBC.dataTypesEnabled` filter directly above use the exact same unguarded `$m['key']` / `$m['provider']` access, and both predate this PR (that filter has been here since 584b3e50). `getTable()` is a thin `apiGet('tables/' . $tableId)` wrapper over a fixed Storage response where each metadata entry is `{id, key, value, provider, timestamp}`. A response malformed enough to break `$m['key']` would already have broken `$tableInfo['isTyped']` and `$tableInfo['rowsCount']` a few lines down, so the new filter is not what exposes it.

On "only drop the specific `storage`-provider description row": the predicate is `$m['key'] !== 'KBC.description' || $m['provider'] !== 'storage'`, so a `KBC.description` row from any other provider survives and still blocks the drop. That is covered both ways in `tests/DeferredTasks/FailedLoadTableDeciderTest.php` — `:65-79` asserts the `storage`-provider row is filtered and the table is dropped, `:80-94` asserts a component-provider row is kept, `count($metadata) !== 0`, and the table is not dropped.

For contrast, `DescriptionHelper::removeDescriptionFromMetadataList()` (`Writer/Helper/DescriptionHelper.php:57-63`) does carry an `is_array()` guard — deliberately, because its input is arbitrary user-supplied pipeline configuration, not a Storage API response.

No change made.

zajca:
@odinuv co ty na to, prošel sem to několikrát a dává mě to smysl, takto.

libs/output-mapping/src/DeferredTasks/LoadTableQueue.php
```diff
@@ -126,6 +160,29 @@ public function waitForAll(): array
         return $jobIds;
     }
 
+    /**
+     * @param array<mixed> $tableData table detail as returned by Storage after a successful load
+     */
+    private function applyCreatedTableDescriptions(LoadTableTaskInterface $task, array $tableData): void
+    {
+        if ($this->createdTableDescriptions === []) {
```
odinuv: tajk ten early exit je tu az zbytecnej, chytil by tohned ten dalsi

libs/output-mapping/src/Mapping/MappingFromConfigurationSchemaColumn.php
```diff
@@ -43,12 +45,34 @@ public function hasMetadata(): bool
...
+     */
+    public function getDescription(): ?string
+    {
         if (isset($this->mapping['description'])) {
-            $metadata['KBC.description'] = $this->mapping['description'];
+            return DescriptionHelper::normalizeDescription($this->mapping['description']);
         }
-        return $metadata;
+
+        // metadata is a variableNode in the configuration, so it is not guaranteed to be an array
```
odinuv: pokud nejsou, tak by to mělo spíš spadnout než to tiše ignorovat (což si myslím, že se někde stane), takže to if (is_array($metadata) se mě moc nelíbí

libs/output-mapping/src/Writer/Helper/DescriptionHelper.php
```diff
@@ -0,0 +1,64 @@
...
+
+    /**
+     * Removes the description from a key => value metadata map.
+     *
+     * @param mixed $metadata the node is a variableNode in the configuration, so it may be anything
+     * @return array<string, mixed>
+     */
+    public static function removeDescriptionFromMetadataMap(mixed $metadata): array
+    {
+        if (!is_array($metadata)) {
```
odinuv: dtto

libs/output-mapping/src/Storage/TableCreator.php
```diff
@@ -5,25 +5,40 @@
...
+     * is always system-managed (Storage default) and there is nothing to diff against.
+     */
     public function createTableDefinition(
         string $bucketId,
         TableDefinitionInterface $tableDefinition,
+        ?TableDescription $descriptions = null,
     ): string {
+        $requestData = $tableDefinition->getRequestData();
+        if ($descriptions !== null) {
+            $requestData = (new CreateTableDefinitionDescriptionEnricher($this->logger))
```
odinuv: nešlo by to to udělat jinak? mě se moc neliíbí, že se tu předávají requestData, celkem dost jsme se snažili, aby, to co se posílá do storage bylo na jednom místě a ne rozfrcaný po X classách, bych si představoval třeba něco jako `$tableDefinition->with($descriptions` který by to nějak pošolíchalo

libs/output-mapping/src/Storage/TableDescriptionModifier.php
```diff
@@ -0,0 +1,116 @@
...
+            return;
+        }
+
+        try {
+            $this->clientWrapper->getTableAndFileStorageClient()->updateTableDefinition(
+                $tableInfo->getId(),
+                $tableDefinitionUpdate,
+            );
+        } catch (ClientException $e) {
+            if ($e->getCode() >= 500) {
```
odinuv: spíš by to mělo být     if ($e->getCode() >= 400 && $e->getCode() < 500) { throw else bubble jinak to nebublá 0

libs/output-mapping/src/Writer/Helper/DescriptionHelper.php
```diff
@@ -0,0 +1,64 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Keboola\OutputMapping\Writer\Helper;
```
odinuv, line 5: Toto nemá testy a kdyby mělo, tak by od oka mohl jít do kélu testDescriptionIsStoredOnCreatedTable kterej je pomalej

libs/output-mapping/tests/Writer/TableDescriptionWriterTest.php
```diff
@@ -0,0 +1,479 @@
...
+        $this->uploadTableWithoutColumns($tableId, 'table description', 'Id description');
+        $this->uploadTableWithoutColumns($tableId, 'table description', 'Id description');
+
+        $tableDetail = $this->clientWrapper->getTableAndFileStorageClient()->getTable($tableId);
+        self::assertSame('table description', $tableDetail['definition']['description'] ?? null);
+        self::assertSame('Id description', $this->getColumnDescription($tableDetail, 'Id'));
+    }
+
+    #[NeedsEmptyOutputBucket]
+    public function testDescriptionIsUpdatedOnTableCreatedByLoadJob(): void
```
odinuv: podle mě nepřináší nic oproti testDescriptionIsUpdatedOnSystemManagedTable a ostatním testům

libs/output-mapping/tests/Writer/TableDefinitionV2Test.php
```diff
@@ -752,6 +752,190 @@ public function testSaveTableAndColumnMetadata(): void
...
+            $this->getColumnDescriptions($tableDetails['definition']['columns']),
+        );
+
+        // Storage rejects a patch without any effective change with 400, so an unchanged third run must not
+        // send one at all
+        $this->uploadTableWithSchema($tableId, 'updated table description', 'updated Id description');
+    }
+
+    #[NeedsEmptyOutputBucket]
+    public function testDescriptionIsNotOverwrittenOnUserManagedTypedTable(): void
```
odinuv: podle mě nepřináší nic oproti testDescriptionIsNotOverwrittenOnUserManagedTable a ostatním testům

libs/output-mapping/tests/Storage/BigqueryTableDescriptionTest.php
```diff
@@ -0,0 +1,146 @@
...
+        self::assertSame('updated table description', $tableDetail['definition']['description'] ?? null);
+        self::assertSame('updated Id description', $this->getColumnDescription($tableDetail, 'Id'));
+    }
+
+    /**
+     * Storage rejects a table-definition patch without any effective change with 400, so an unchanged repeated
+     * run must not send one.
+     */
+    #[NeedsEmptyBigqueryOutputBucket]
+    public function testRepeatedRunWithUnchangedDescriptionSucceeds(): void
```
odinuv: nevypadá, že by to bylo bigquery specific

libs/output-mapping/tests/Storage/BigqueryTableDescriptionTest.php
```diff
@@ -0,0 +1,146 @@
...
+
+        $this->uploadTable($tableId, 'table description', 'Id description');
+
+        $tableDetail = $this->clientWrapper->getTableAndFileStorageClient()->getTable($tableId);
+        self::assertSame('table description', $tableDetail['definition']['description'] ?? null);
+        self::assertSame('Id description', $this->getColumnDescription($tableDetail, 'Id'));
+    }
+
+    #[NeedsEmptyBigqueryOutputBucket]
+    public function testDescriptionIsUpdatedOnExistingTable(): void
```
odinuv: vlastně ani tohle ne, nepřijde mi, že je tu nějaký rozdíl mezi BQ a Snowflakem a jestli není tady, tak ho tady nechcem testovat (věříme tomu, že je to otestované na connection)

odinuv: případně by se to dalo řešit přídáním assertions v TableDefinitionV2BigQueryTest ?

libs/output-mapping/tests/Writer/TableDefinitionTest.php
```diff
@@ -688,4 +688,111 @@ private static function assertTablePrimaryKeyAddJob(array $jobData, array $expec
...
         self::assertSame($expectedPk, $jobData['operationParams']['columns']);
     }
+
+    /**
+     * The legacy native-types path builds the table definition from `column_metadata`, and the descriptions
+     * ride along in the create payload. A following run finds the table there and goes through the
+     * table-definition update instead.
+     */
+    #[NeedsEmptyOutputBucket]
+    public function testDescriptionIsStoredAndUpdatedOnTypedTable(): void
```
odinuv: afaik tady by taky stačilo přidat description a assert do testWriterUpdateTableDefinitionWithBaseTypes

libs/output-mapping/tests/Writer/Workspace/WriterWorkspaceTest.php
```diff
@@ -99,6 +99,64 @@ public function testSnowflakeTableOutputMapping(): void
         );
     }
 
+    /**
+     * The descriptions are stored through the table-definition API regardless of the staging the data came
+     * from, but workspace unload reaches LoadTableTaskCreator through a different strategy.
+     */
+    #[NeedsTestTables(2), NeedsEmptyOutputBucket]
+    public function testSnowflakeTableOutputMappingStoresDescription(): void
```
odinuv: dtto

libs/output-mapping/tests/Writer/TableDefinitionV2Test.php
```diff
@@ -752,6 +752,190 @@ public function testSaveTableAndColumnMetadata(): void
...
+        self::assertSame('storage', array_values($filteredColumnNameMetadata)[0]['provider']);
+    }
+
+    /**
+     * On the second run the table already exists, so the description goes through StoragePreparer, which diffs
+     * it against the value stored in the table definition. A typed table exposes it in the same place a
+     * non-typed one does, otherwise the diff would keep re-sending an unchanged patch.
+     */
+    #[NeedsEmptyOutputBucket]
+    public function testDescriptionIsUpdatedOnExistingTypedTable(): void
```
odinuv: dtto

odinuv (commented):
V zásadě je to podle mě dobrý, mám tam pár drobností k implementaci, viz commenty

PR description není úplně aktuální - tohle - , created by the load job itself), and embedding would break FailedLoadTableDecider::decideTableDelete() — a description supplied at create time makes Storage write a metadata row immediately, so count($metadata) === 0 would no longer hold and empty tables would stop being dropped after a failed import.

ale hlavně prosím zredukovat funkční testy na minimum, tohle natáhlo testy o 3 minuty a řekl bych, že z velké části zbytečně, imho stačí přidat assert tamhle a onamhle a bude to testovat to samý

libs/output-mapping/src/Storage/TableCreator.php
odinuv: co hlavně nechcu, aby v dalším PR někdo šel a začal s tím $requestData dělat nějakou další chytristiku, protože pak to znamená, že  $tableDefinition->getRequestData(); vrací něco nereálnýho a celej ten systém se zhroutí :)

libs/output-mapping/src/DeferredTasks/LoadTableQueue.php
zajca: Pryč.

libs/output-mapping/src/Mapping/MappingFromConfigurationSchemaColumn.php
zajca: Máš pravdu a bylo to horší, než to vypadá — na mainu se v tom případě spadne, takže jsem tím tichým ignorováním zaváděl regresi.

Teď je to jeden throw v `DescriptionHelper` a obě čtení jdou přes něj. Dva testy, co pinovaly to staré chování, jsem přepsal na `expectException`.

libs/output-mapping/src/Writer/Helper/DescriptionHelper.php
zajca: Vyřešeno spolu s tím druhým — obojí teď jde přes jeden throw v helperu.

libs/output-mapping/src/Storage/TableCreator.php
zajca: Souhlas, přepsáno. `TableDefinitionInterface` dostalo `setDescriptions()` a popisky si každá implementace vyrenderuje uvnitř svého `getRequestData()`. Enricher je smazaný.

Bonusem se `TableCreator` vrátil na původní signaturu, takže padly i dvě BC připomínky od review bota.

libs/output-mapping/src/Storage/TableDescriptionModifier.php
zajca: Opraveno podle tebe, wrapuje se jen 4xx.

libs/output-mapping/src/Writer/Helper/DescriptionHelper.php
zajca: Přidán unit test, 28 případů. A ano — díky němu šel `testDescriptionIsStoredOnCreatedTable` z 3 běhů přes dataProvider na 1.

libs/output-mapping/tests/Writer/TableDescriptionWriterTest.php
zajca: Smazáno.

Navíc jsem posloučil dvojice test + „a ještě jednou totéž" do jednoho testu s druhým uploadem uvnitř, a vyhodil `testFailedLoadKeepsPreExistingTableWithDescription` — to je původní chování `FailedLoadTableDecider`, s popisky nesouvisí.

libs/output-mapping/tests/Writer/TableDefinitionV2Test.php
zajca: Smazáno — user-managed gate na typovanosti nezávisí, takže duplicita.

libs/output-mapping/tests/Storage/BigqueryTableDescriptionTest.php
zajca: Souhlas — celý soubor je pryč.

libs/output-mapping/tests/Storage/BigqueryTableDescriptionTest.php
zajca: Souhlasím, žádný rozdíl proti Snowflaku tam není. Smazáno.

zajca: Vzal jsem tenhle návrh — asserty jsou v `TableDefinitionV2BigQueryTest::testWriteTableOutputMappingExistingTable`, který stejně nahrává dvakrát, takže první běh ověří create payload a druhý to, že se neposílá nezměněný patch.

libs/output-mapping/tests/Writer/TableDefinitionTest.php
zajca: Smazáno — create asserty jsou v `testWriterCreateTableDefinition`, update v `testWriterUpdateTableDefinitionWithBaseTypes`. Ale odhalilo to jednu věc, kterou bys měl vidět.

Přidání popisku rozbilo assert, co s popisky vůbec nesouvisí: `created` hlásil `TIMESTAMP_LTZ` místo `TIMESTAMP`. Důvod je v connection — description-only zkratka v `TableDefinitionUpdateService` je gatovaná na netypovanou tabulku, takže u typované jde i čistě popiskový patch cestou, která přečte všechny sloupce z backendu a přepíše jim datatype metadata. Je to no-op tam, kde metadata sedí, proto se ostatní sloupce nehnuly. `created` se přidával přes basetype a tam si Storage poznamená požadovaný alias `TIMESTAMP`, i když sloupec je fyzicky `TIMESTAMP_LTZ` s délkou 9. Metadata byla špatně už od přidání sloupce, jen to do teď nic nevyvolalo.

Očekávání v testu jsem tedy srovnal s realitou. Popis jen na úrovni tabulky resync nespustí a netypované tabulky taky ne, takže dopad je omezený na typované tabulky, kde ukládáme popis sloupce.

Ten `tableColumnAdd` je samostatná chyba, s tímhle PR nesouvisí — založil jsem na ni [DMD-1849](https://linear.app/keboola/issue/DMD-1849).

libs/output-mapping/tests/Writer/Workspace/WriterWorkspaceTest.php
zajca: Smazáno, asserty doplněné do `testSnowflakeTableOutputMapping`. Ušetří to celý `initWorkspace()`.

libs/output-mapping/tests/Writer/TableDefinitionV2Test.php
zajca: Taky smazáno. Update cestu i popis sloupce přidaného stejným během pokryje `testAddMissingColumnTableDefinition`, stačilo tam doplnit popisky a asserty. Nula uploadů navíc.

libs/output-mapping/src/Storage/TableCreator.php
zajca: Přesně proto to skončilo uvnitř `getRequestData()` a ne za ním — vrací teď doslova to, co jde do Storage, takže není kam tu chytristiku narvat.

zajca:
@odinuv díky, prošel jsem všechno — odpovědi jsou u jednotlivých komentářů.

Testy zredukované: vyhrazené běhy proti Storage z 22 na 7, všechny zbylé v `TableDescriptionWriterTest`. Zbytek pokrytí je doplněný jako asserty do testů, které ty cesty projíždějí tak jako tak, takže nestojí ani jeden load job navíc. `BigqueryTableDescriptionTest` je celý pryč.

Popis PR aktualizovaný.

Jedna věc navíc, kterou to vyhrabalo — je to u komentáře k `TableDefinitionTest`: uložení popisu sloupce na **typované** tabulce donutí Storage přečíst všechny sloupce z backendu a přepsat jim datatype metadata. Je to no-op tam, kde metadata sedí, ale u sloupce přidaného přes basetype se tím opraví hodnota, která tam byla špatně už od jeho přidání (`TIMESTAMP` → `TIMESTAMP_LTZ`). Očekávání v testu jsem srovnal s realitou. Ten `tableColumnAdd` je samostatná chyba, na kterou jsem založil [DMD-1849](https://linear.app/keboola/issue/DMD-1849).

CI zelená.

odinuv (approved):
👍

PR #551, opened 2026-07-27
refactor(output-mapping): create non-typed tables through the table-definition endpoint

zajca:
Preparation for [AJDA-2946](https://linear.app/keboola/issue/AJDA-2946/output-mapping-store-description-only-when-table-flag-is-system) (#550), useful on its own.

## Why

Output mapping has two ways of creating a table. Typed tables and tables declared by `schema` go through `createTableDefinition`. A table declared by manifest `columns` without data types went through `createTableAsync`, which requires a header CSV to be written to a temp file and uploaded to file storage before the table can be created.

The definition endpoint can create that table too: Storage reports a table as typed only when at least one column carries `definition.type` or `basetype` (`BaseCreateTableDefinitionRequest::isTypedTable()`), so a payload with plain column names produces a non-typed table. Verified against real Storage — same `isTyped=false`, same columns, same primary key, data loads normally.

So this switches the untyped path to `createTableDefinition`, which:
- drops a file upload and a storage job from every non-typed table creation,
- leaves output mapping with a single way of creating a table,
- lets the description ride along in the create payload later (that is #550's business, not this PR's).

## Changes

- new `TableDefinitionFromColumns` — builds `{name, primaryKeysNames, columns: [{name}]}` and nothing else, so no column can accidentally acquire a type
- `LoadTableTaskCreator` uses it for the `!didTableExistBefore() && hasColumns()` branch
- `TableCreator::createTable()` removed — this change made it unused, and with it the header-CSV plumbing (`Temp`, `CsvFile`)

## Observable differences

**`distributionKey` is no longer set on the created table.** Measured side by side, this is the only field that differs; everything else (isTyped, primaryKey, columns, column definitions, metadata rows, import) is identical.

That is acceptable because Storage documents the field as *"Distribution key columns (for Synapse)"*, Synapse is no longer a Storage backend, and on Snowflake it is inert metadata echoed back in the table detail. Note the typed and schema paths already do not set it — so today a typed table created from a config with `distribution_key` gets no distribution key while an untyped one does. This makes the two consistent.

**An invalid column name is reported differently.** `createTableAsync` said `'<name>' is more than 64 characters long`; the definition endpoint says `Invalid column name: "<name>"`. The new message does not spell out the limit. `testLongColumnName` was updated accordingly.

## Testing

- phpcs + phpstan (level max) clean.
- `TableCreatorTest` — the two tests covering the old `createTable()` were converted to the new path rather than deleted, so non-typed creation with a primary key and its error handling stay covered against real Storage.
- New unit test for `TableDefinitionFromColumns`, including that a sparse column list (RestrictedColumnsHelper filters with `array_filter` and leaves gaps) still serialises as a JSON array.
- `main-writer-tests-1` run against real Storage: the only failure caused by this change was `testLongColumnName` (message text, updated). The four `Can not resolve branchId` dev-branch errors are pre-existing — verified by re-running one of them with these changes stashed, where it fails identically.

### DIFF
--- a/libs/output-mapping/tests/Writer/StorageApiLocalTableWriterTest.php
+++ b/libs/output-mapping/tests/Writer/StorageApiLocalTableWriterTest.php
@@ -2275,9 +2275,11 @@ CSV;
             );
             $this->fail('Must throw exception');
         } catch (InvalidOutputException $e) {
+            // the table is created through the table-definition endpoint, which reports an invalid column
+            // name without spelling out the length limit
             $this->assertStringContainsString(
-                '\'LongLongLongLongLongLongLongLongLongLongLongLongLongLongLongLongName\' '.
-                'is more than 64 characters long',
+                'Invalid column name: '.
+                '\\"LongLongLongLongLongLongLongLongLongLongLongLongLongLongLongLongName\\"',
                 $e->getMessage(),
             );
         }
--- a/libs/output-mapping/src/LoadTableTaskCreator.php
+++ b/libs/output-mapping/src/LoadTableTaskCreator.php
@@ -14,6 +14,7 @@ use Keboola\OutputMapping\Storage\NativeTypeDecisionHelper;
 use Keboola\OutputMapping\Storage\TableCreator;
 use Keboola\OutputMapping\Writer\Table\StrategyInterface;
 use Keboola\OutputMapping\Writer\Table\TableDefinition\TableDefinitionFactory;
+use Keboola\OutputMapping\Writer\Table\TableDefinition\TableDefinitionFromColumns;
 use Keboola\OutputMapping\Writer\Table\TableDefinitionFromSchema\TableDefinitionFromSchema;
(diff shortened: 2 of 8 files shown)


libs/output-mapping/tests/Writer/StorageApiLocalTableWriterTest.php
```diff
@@ -2275,9 +2275,11 @@ public function testLongColumnName(): void
             );
             $this->fail('Must throw exception');
         } catch (InvalidOutputException $e) {
+            // the table is created through the table-definition endpoint, which reports an invalid column
+            // name without spelling out the length limit
```
odinuv, line 2279: to je škoda :)

odinuv (approved):
může bejt

podle mě ty komentáře, který popisují změnu "// takže vznikne netypovaná tabulka stejně jako dřív, ale bez uploadu hlavičkového CSV souboru" všeobecně nepatří do kódu, ale do PR

PR #553, opened 2026-07-30
fix(output-mapping): wait for the direct-grant metadata refresh job (DMD-1828)

zajca:
**Linear**: DMD-1828
Other repositories: keboola/connection#7992, keboola/connection#7993

This is **PR 3/3** of the DMD-1828

## What changed and why

Customer symptom (SUPPORT-17181): a Flow trigger on a table written by direct-grant output mapping fired only intermittently, although the transformation succeeded on every scheduled run.

Direct-grant output mapping does not create a `tableImport` job — `TableLoader` calls
`POST /workspaces/{id}/unload?only-direct-grants=true`, and the async `refreshStorageBuckets` job
that returns is what refreshes table metadata and dispatches the `TABLE_IMPORT_DONE` events Flow
triggers listen to. That call threw the returned job ids away and swallowed every `Throwable` into
a log warning, so:

- nothing waited for the refresh, and `docker-bundle` drops the staging workspace right after
  output mapping — so the refresh could reach its workspace lookup after the workspace was already
  deleted, fail, and never dispatch an event;
- that failure was invisible: the transformation reported Success with stale metadata and an
  untriggered flow.

The refresh is now a task in `LoadTableQueue` like any other Storage job, which closes the race by
construction: `Runner::waitForStorageJobs()` calls `waitForAll()` first and only cleans up the
staging workspace in its `finally` block. An errored refresh job lands in the same `$errors` list
as a failed table load, i.e. in the single `InvalidOutputException`. Enqueuing the refresh no
longer swallows failures either: a 4xx becomes an `InvalidOutputException`, anything else is
rethrown.

## How it is wired

`DeferredTaskInterface` describes what the queue needs from any Storage job it awaits — `start()`,
`getStorageJobIds()` and `getFailedJobError()`. `LoadTableTaskInterface` extends it with the
table-load specifics (metadata, destination table, freshly-created flag). The refresh is
`DirectGrantMetadataRefreshTask`, first in the same task list, so it is enqueued by
`LoadTableQueue::start()` and awaited by the same loop as the table loads. It goes first because
`start()` walks the list and a failing enqueue aborts it — the direct-grant tables are already
written at that point, so their metadata has to catch up even when a table load is rejected. A task
without a destination table only contributes its error message — it has nothing to annotate,
describe or measure, so `refreshStorageBuckets` never reaches `$jobResults` / `Metrics` /
`TableInfo`.

Mapping a Storage failure of the enqueue call to `InvalidOutputException` moved from
`LoadTableQueue::start()` into the tasks, because the two differ: a table load treats any code
`< 500` as the user's fault (unchanged behaviour), while the refresh treats only a 4xx that way.
`ClientException` carries code `0` when there is no HTTP response (connection failure, DNS, read
timeout) or when the client throws before the request, and those must not be reported as the
user's fault. Fixing the table-load condition changes the behaviour of an existing path and
belongs in its own PR.

## BEHAVIOUR CHANGE

`docker-bundle` converts `InvalidOutputException` to a `UserException`, so **transformations using
`unload_strategy: direct-grant` will now FAIL when the Storage metadata refresh fails**, where they
previously reported Success. Enqueuing the refresh can also fail the job outright (4xx) instead of
logging a warning.

**This must be released after the two connection-side fixes for DMD-1828 land**
(per-table tolerance of a missing backend object, and bucket-scoped locking for the refresh job).
Today's dominant refresh failures are exactly the ones those PRs remove — project 15 reproduces
hourly — so releasing this first would convert existing silent failures into hard transformation
failures. Note also that `Client::waitForJob()` polls without a timeout, so before those fixes a
starved refresh would additionally block the transformation for minutes.

## API changes in the published library

- `LoadTableTaskInterface`: `startImport(Client)` → `start(ClientWrapper)` (a task now picks the
  client it needs — the refresh has to go through the branch client), and
  `getStorageJobId(): string` → `getStorageJobIds(): array`. `POST /workspaces/{id}/unload` is typed
  as a list of jobs: today it enqueues a single `refreshStorageBuckets` job carrying every
  direct-grant bucket of the workspace (bucket-scoped locking in keboola/connection#7993 splits the
  locks, not the job), it returns an empty list when the workspace configuration has no direct-grant
  output table, and it will grow the jobs of the remaining unload strategies. Only
  `AbstractLoadTableTask` implements the interface.
- `LoadTableQueue::waitForAll()` returns the ids of **all** Storage jobs it waited for, not just
  the table load ones. No production consumer reads the return value — `docker-bundle`
  `Runner.php:407`, sandboxes `WorkspaceManager.php:248` and data-loader-api `DataExporter.php:96`
  all discard it.
- `LoadTableQueue::getTaskCount()` counts Storage jobs, so `docker-bundle`'s
  "Waiting for N Storage jobs to finish." is truthful; a pure direct-grant run used to log `0` and
  then wait for one job. It is only meaningful once the queue has been started.
- `LoadTableQueue::getLoadTableTasks()` still returns table load tasks only.

## Testing

- `tests/DeferredTasks/LoadTableQueueTest.php` — the queue now delegates the error message to the
  task; new tests for a task without a destination table (awaited, excluded from metrics, its
  error surfaces), for `getTaskCount()` over every job id of a task, and for `start()` enqueuing in
  list order up to the first failure (what puts the refresh first).
- `tests/DeferredTasks/DirectGrantMetadataRefreshTaskTest.php` — the "not started yet" guard and an
  empty job list as a valid result of the unload call.
- `tests/DeferredTasks/LoadTableTaskTest.php` — the enqueue error mapping (4xx →
  `InvalidOutputException` with code and previous preserved, 5xx rethrown), the "not started yet"
  guard, and the failed-job message.
- `tests/TableLoaderDirectGrantMetadataRefreshTest.php` (renamed from
  `TableLoaderDirectGrantUnloadTest`) — 6 cases driving `TableLoader::uploadTables()`: the refresh
  is enqueued with `only-direct-grants`, every returned job id is awaited, an errored job surfaces,
  a 404 becomes `InvalidOutputException`, a 500 and a connection failure (code `0`) are both
  rethrown as `ClientException`.
- `tests/Writer/Workspace/TableLoaderUnloadStrategyTest.php` — `testDirectGrantMetadataRefreshIsAwaited`
  really goes through the direct-grant path: the refresh job is the only queued job, it is a
  `refreshStorageBuckets` job and the queue waits for it. Storage resolves the direct-grant output
  tables of a workspace from the component configuration the workspace was created from
  (`WorkspaceUnloadProcessor::process()`), so a workspace created by `Workspaces::createWorkspace()` —
  what `initWorkspace()` does — always gets an empty list of jobs back and can never produce a
  refresh job; that is why the existing tests in this file only ever asserted that a direct-grant
  mapping uploads nothing. The new `AbstractTestCase::initConfigurationWorkspace()` creates a
  `keboola.snowflake-transformation` configuration with a direct-grant output table and the workspace
  from it.
  Asserting that the table itself ends up registered in Storage is out of reach for a test:
  `/workspaces/{id}/query` restricts every statement to the workspace schema
  (`ExecuteQueryProcessor` passes `pathRestriction`), so the test cannot write the table into the
  bucket schema the way a transformation holding the direct grant does.
- `testDirectGrantUnloadStrategySkipsTableUpload` removed from `TableDefinitionV2Test`: it belonged
  to table definitions, duplicated `testDirectGrantSkipsTableImport`, and its job-count assertion
  held for a workspace which had nothing to unload in the first place.
- A now-impossible log-based check in `TableLoaderUnloadStrategyTest` replaced by a task-count
  assertion with the same intent.
- `vendor/bin/phpunit tests/DeferredTasks` → **54 tests, 201 assertions, OK**;
  `vendor/bin/phpunit tests/TableLoaderDirectGrantMetadataRefreshTest.php` → **6 tests, 42
  assertions, OK**. `composer check` (validate + phpcs + PHPStan level max) → clean.
- Functional suites run in CI. `Workspace writer tests` cover the new direct-grant test.
- **Unrelated red CI:** `General tests`
  (`TableStructureModifierTest::testUpdateTableStructureAddColumnsFromMetadata`) and `Native types`
  (`TableDefinitionTest::testWriterUpdateTableDefinitionWithNativeTypes`) fail on column types
  Storage now returns — `INTEGER` → `NUMBER` with `length: 38,0`, `DECIMAL` → `NUMBER`. Both
  reproduce on the branch point without any of the changes in this PR, so they need their own fix
  before `tests-result` can go green.

### DIFF
--- a/libs/output-mapping/src/DeferredTasks/LoadTableQueue.php
+++ b/libs/output-mapping/src/DeferredTasks/LoadTableQueue.php
@@ -27,22 +27,27 @@ class LoadTableQueue
     /** @var array<string, TableDescription> table id => descriptions of a table created by this run */
     private array $createdTableDescriptions;
 
+    /** @var int[] */
+    private array $unloadJobIds;
     private Result $tableResult;
 
     /**
      * @param LoadTableTaskInterface[] $loadTableTasks
      * @param array<string, TableDescription> $createdTableDescriptions
+     * @param int[] $unloadJobIds Storage metadata refresh jobs enqueued for direct-grant tables
      */
     public function __construct(
         ClientWrapper $clientWrapper,
         LoggerInterface $logger,
         array $loadTableTasks,
         array $createdTableDescriptions = [],
+        array $unloadJobIds = [],
     ) {
         $this->clientWrapper = $clientWrapper;
         $this->logger = $logger;
         $this->loadTableTasks = $loadTableTasks;
         $this->createdTableDescriptions = $createdTableDescriptions;
+        $this->unloadJobIds = $unloadJobIds;
         $this->tableResult = new Result();
     }
 
@@ -65,6 +70,9 @@ class LoadTableQueue
         }
     }
 
+    /**
+     * @return string[] ids of the table load jobs
+     */
     public function waitForAll(): array
     {
         $metadataApiClient = new Metadata($this->clientWrapper->getTableAndFileStorageClient());
@@ -154,6 +162,20 @@ class LoadTableQueue
 
         $this->tableResult->setMetrics(new Metrics($jobResults));
 
+        // the metadata refresh dispatches trigger events; it must finish before the workspace is dropped
+        foreach ($this->unloadJobIds as $unloadJobId) {
+            /** @var array $unloadJobResult */
+            $unloadJobResult = $this->clientWrapper->getBranchClient()->waitForJob($unloadJobId);
+
+            if ($unloadJobResult['status'] === 'error') {
+                $errors[] = sprintf(
+                    'Failed to refresh metadata of direct-grant tables (Storage job "%s"): %s',
+                    $unloadJobId,
+                    $unloadJobResult['error']['message'],
+                );
+            }
+        }
+
         if ($errors) {
             throw new InvalidOutputException(implode("\n", $errors));
         }
@@ -179,9 +201,12 @@ class LoadTableQueue
         $descriptionModifier->updateDescriptions(new StorageTableInfo($tableData), $descriptions);
     }
 
+    /**
+     * Number of Storage jobs waitForAll() waits for.
+     */
     public function getTaskCount(): int
     {
-        return count($this->loadTableTasks);
+        return count($this->loadTableTasks) + count($this->unloadJobIds);
     }
 
     public function getTableResult(): Result
--- a/libs/output-mapping/src/DeferredTasks/LoadTableQueue.php
+++ b/libs/output-mapping/src/DeferredTasks/LoadTableQueue.php
@@ -21,57 +21,40 @@ class LoadTableQueue
     private ClientWrapper $clientWrapper;
     private LoggerInterface $logger;
 
-    /** @var LoadTableTaskInterface[] */
-    private array $loadTableTasks;
+    /** @var DeferredTaskInterface[] */
+    private array $tasks;
 
     /** @var array<string, TableDescription> table id => descriptions of a table created by this run */
     private array $createdTableDescriptions;
 
-    /** @var int[] */
-    private array $unloadJobIds;
     private Result $tableResult;
 
     /**
-     * @param LoadTableTaskInterface[] $loadTableTasks
+     * @param DeferredTaskInterface[] $tasks
      * @param array<string, TableDescription> $createdTableDescriptions
-     * @param int[] $unloadJobIds Storage metadata refresh jobs enqueued for direct-grant tables
      */
     public function __construct(
         ClientWrapper $clientWrapper,
         LoggerInterface $logger,
-        array $loadTableTasks,
+        array $tasks,
         array $createdTableDescriptions = [],
-        array $unloadJobIds = [],
     ) {
         $this->clientWrapper = $clientWrapper;
         $this->logger = $logger;
-        $this->loadTableTasks = $loadTableTasks;
+        $this->tasks = $tasks;
         $this->createdTableDescriptions = $createdTableDescriptions;
-        $this->unloadJobIds = $unloadJobIds;
         $this->tableResult = new Result();
     }
 
     public function start(): void
     {
-        foreach ($this->loadTableTasks as $loadTableTask) {
-            try {
-                $loadTableTask->startImport($this->clientWrapper->getTableAndFileStorageClient());
-            } catch (ClientException $e) {
-                if ($e->getCode() < 500) {
-                    throw new InvalidOutputException(
-                        sprintf('%s [%s]', $e->getMessage(), $loadTableTask->getDestinationTableName()),
-                        $e->getCode(),
-                        $e,
-                    );
-                }
-
-                throw $e;
-            }
+        foreach ($this->tasks as $task) {
+            $task->start($this->clientWrapper);
         }
     }
 
     /**
-     * @return string[] ids of the table load jobs
+     * @return string[] ids of the Storage jobs waited for
      */
     public function waitForAll(): array
     {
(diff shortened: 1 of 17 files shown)


libs/output-mapping/src/TableLoader.php
```diff
@@ -165,10 +167,10 @@ public function uploadTables(
                 ));
             }
             // enqueue unload for direct-grant tables
-            $this->callWorkspaceUnload($strategy);
```
odinuv, line 182: A nemohlo by to tady zakládat nějakej LoadTableTaskInterface normálně do fronty jako ostatní loady? To je přece jedno, co ten job ve výsledku udělá, z pohledu OM je to job jako každej jinej (možná osekanej). 
Spustilo by se to tím LoadTableQueue->start jako ostatní tasky, tady na pořadí nijak nezáleží ne?

Kromě toho je zmatený to pojmenování load/unload, to zní jakoby to dělalo něco opačnýho a přitom to obojí "nahrává" data do storage

libs/output-mapping/src/DeferredTasks/LoadTableQueue.php
```diff
@@ -120,15 +132,32 @@ public function waitForAll(): array
 
         $this->tableResult->setMetrics(new Metrics($jobResults));
 
+        // the metadata refresh dispatches trigger events; it must finish before the workspace is dropped
+        foreach ($this->unloadJobIds as $unloadJobId) {
```
odinuv: taky taky tím, že je to oddělený, to chybí v jobResults nahoře, přijde mi, že k tomu není moc důvod (kromě toho, že to tak jako už bylo)

libs/output-mapping/src/TableLoader.php
zajca: ok akorát ten interface musím rozšířit bude vracat int[] místo int

odinuv: Jakto? $workspaces->queueUnload( vrací jeden job ne?

zajca: Nene pole storage jobů, pro každý bucket zvlášť refresh job

zajca:
Co takto?

libs/output-mapping/src/TableLoader.php
```diff
@@ -178,8 +178,10 @@ public function uploadTables(
                     $strategy::class,
                 ));
             }
-            // enqueue unload for direct-grant tables
-            $this->callWorkspaceUnload($strategy);
+
+            $loadTableTasks[] = new DirectGrantMetadataRefreshTask(
```
odinuv: rename tady jeste

libs/output-mapping/tests/Writer/TableDefinitionV2Test.php
```diff
@@ -1124,7 +1124,7 @@ public function testDirectGrantUnloadStrategySkipsTableUpload(): void
             systemMetadata: new SystemMetadata(['componentId' => 'foo']),
         );
         $jobIds = $tableQueue->waitForAll();
-        self::assertCount(0, $jobIds, 'No jobs should be created when unload_strategy is direct-grant');
+        self::assertCount(0, $jobIds, 'No table load job should be created when unload_strategy is direct-grant');
```
odinuv: to by ale nemělo bejt 0 jobů ne? waitForAll vrací všechny (což je dobře)

libs/output-mapping/src/DeferredTasks/TableWriter/AbstractLoadTableTask.php
```diff
@@ -47,9 +71,25 @@ public function getDestinationTableName(): string
         return $this->destination->getTableId();
     }
 
-    public function getStorageJobId(): string
+    public function getStorageJobIds(): array
+    {
+        if (!isset($this->storageJobId)) {
```
odinuv, line 76: akorátže DirectGrantMetadata vrací prázdný pole a tuhle výjimku nevyhazuje https://github.com/keboola/platform-libraries/pull/553/changes#diff-761a54194784a5b2f2295e1963d029602c34270e8597ca6a0e7e6ba40e05416aR52 
reálně se to asi nestane, ale mělo by se to chovat stejně ať tak nebo tak

libs/output-mapping/tests/Writer/TableDefinitionV2Test.php
odinuv: No protože ten test vůbec nic netestuje - ta tabulka v tom WS není, takže se nic neunloadne 🤦

odinuv (approved):
vpodstatě ok, hlavně prosím oprav ten test testDirectGrantUnloadStrategySkipsTableUpload ať teda opravdu testuje ten directgrant , pak se to může mergnout

zajca:
Pofixoval sem to, ale ještě musím předto hodit PR které opravý očekávání z connection protože sem fixoval add column endpoint a padajů teď testy

PR #554, opened 2026-08-03
feat(api-bundle): Support form payloads and per-property wire keys

pepamartinec:
## Release Notes

Adds two capabilities to the shared `RequestMapper` so services no longer have to fork it. Both come out of `oauth-service`, which currently ships a local copy of the bundle's resolver (`App\RequestMapper`) purely to serve two endpoints the shared one cannot: a form-encoded `POST /authorize` and a credentials payload whose encrypted field arrives under the wire key `#data`. With this change that local copy can be deleted.

`#[RequestPayloadObject]` gains a `format` option (`PayloadFormat::Json` by default, so nothing changes for existing callers). A `Form` payload is read from the request parameters rather than the raw body and mapped with casting enabled, because form values are always strings; a mismatched content type is rejected with `406` exactly as the JSON path already does. Separately, `#[RequestKey('#data')]` is declared **on the property itself**, so the wire contract sits next to the field it describes rather than being repeated at every call site. Two details are deliberate: validation errors are reported under the wire key (`#data`), not the property name, so the client sees the field it actually sent; and a value posted directly under the property name is dropped, so the result cannot depend on the order of keys in the request. Two properties claiming the same wire key throw `RequestMapperException`.

Alternatives considered:
* **Valinor 2 `#[AsConverter]` + the documented `RenameKeys` converter** — the natural fit on paper, but it reports errors under the property name (`data`), passes unmapped keys straight through (so a bare `data` still wins, order-dependently), and would require bumping `cuyz/valinor-bundle` `^0.4` → `^2.3`. That upgrade removes `MapperBuilder::enableFlexibleCasting()`, `Messages::flattenFromNode()` and `NodeMessage::node()` — all used by `DataMapper` — putting every consumer's query mapping in the blast radius.
* **A named constructor taking a shaped array with a quoted `'#data'` key** — reports the true wire key, but a missing body collapses into a single error on the constructor parameter instead of one error per field, which defeats the point of per-field reporting. Verified against both Valinor 1.17 and 2.5.1.

Keeping the mapping in the bundle avoids the upgrade entirely and gets a strictly better result than either.

None.

## Impact analysis
No impact on existing consumers. `format` defaults to `PayloadFormat::Json`, the JSON content-type error message is unchanged, and the key map only does anything on classes carrying `#[RequestKey]`. The added per-class reflection lookup is memoized and measures ~0.3 µs against ~71 µs for a single `map()` call.

## Change type
New feature

## Justification
Lets `oauth-service` drop its forked request mapper; both features are generic enough that the next service to hit either case won't fork it again.

## Deployment
Merge & release a new bundle version.

## Rollback plan
Revert of this PR.

## Post release support plan
None.

### DIFF
--- a/libs/api-bundle/README.md
+++ b/libs/api-bundle/README.md
@@ -145,6 +145,64 @@ which the bundle requires directly, so no extra installation is needed.
 > If you forget to install appropriate client, you will get exception like
 > `Service "Keboola\ApiBundle\Attribute\ApplicationTokenAuth" not found: the container inside "Symfony\Component\DependencyInjection\Argument\ServiceLocator" is a smaller service locator`
 
+## Request mapping
+
+Type-hint a payload object on a controller argument and annotate it with
+`#[RequestPayloadObject]` (request body) or `#[RequestQueryObject]` (query string). The bundle maps
+and validates the request with Valinor + the Symfony validator, and reports every failing field
+individually in the error `context` as `{path, message}`.
+
+```php
+public function __invoke(
+    #[RequestPayloadObject] CreateThingPayload $payload,
+    #[RequestQueryObject] ListFilters $filters,
+): Response {
+}
+```
+
+Both attributes accept `allowExtraKeys` (default `true`) to control whether unknown keys are
+rejected.
+
+### Form-encoded payloads
+
+`#[RequestPayloadObject]` expects `application/json` by default. Endpoints that receive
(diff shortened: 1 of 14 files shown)


pepamartinec:
@janvanicek co tohle misto https://github.com/keboola/oauth-service/pull/86?

janvanicek (approved):
Pusťme to takhle ven

PR #555, opened 2026-08-04
fix(input-mapping): report input file download failures as user errors (SUPPORT-17151)

odinuv:
## Description

Python transformation (and other component) jobs sometimes failed with a generic **"Internal Server Error occurred."** when an input **file** could not be downloaded from Storage — even though the underlying cause is user-owned (the file is missing or has expired). See SUPPORT-17151 (Datadog: the failing job's last S3 GET returned `404 NoSuchKey` for an input file).

Root cause: `File\Strategy\AbstractStrategy::downloadFiles()` caught **every** `Throwable` from the per-file download and re-wrapped it as an `InputOperationException`, which the consumer (`job-queue`'s `InputDataLoader`) treats as an application error. The Storage client already throws a clear, user-facing message for this case (`Cannot download file "…" (ID …) from Storage, please verify the contents of the file and that the file has not expired.`), but it was masked.

Fix: classify Storage `ClientException`s during file download as a user error (`InvalidInputException`, caught downstream and surfaced as a `UserException`), while keeping genuinely unexpected errors as `InputOperationException`. The job now ends as a **user error** with the actionable Storage message instead of a masked internal error.

Note: despite the name, `InputMapping\Exception\FileNotFoundException` is **not** raised for a missing/expired Storage file — it is an internal control-flow signal thrown by `InputFileStateList::getFile()` when no incremental-load *state entry* matches the file tags, and is caught locally to fall back to empty state. A file that cannot be downloaded surfaces as a Storage `ClientException`, which is what this change now classifies. https://github.com/keboola/connection/blob/5e1f562d7d4c11df964ab249a49e309267acb127/Package/StorageApiPhpClient/src/Keboola/StorageApi/Client.php#L2518

Transported from keboola/input-mapping#2 into the monorepo; the test is folded into the existing `DownloadFilesTest` rather than a dedicated class.

## Justification

See [SUPPORT-17151](https://keboola.atlassian.net/browse/SUPPORT-17151) — missing/expired input files were reported as internal errors instead of actionable user errors.

None.

## Impact Analysis

Affects error classification only for the file input-mapping download path. Failures previously reported as application/internal errors for missing/expired input files will now be reported as user errors. No change to successful runs or to genuinely unexpected errors. Not feature-flagged.

## Deployment Plan

Release a tagged version (`input-mapping/<version>`) after merge. And update IM in job-runner (to fix python transformations, the rest will update organically)

## Rollback Plan

Revert this PR.

## Post-Release Support Plan

None.

[SUPPORT-17151]: https://keboola.atlassian.net/browse/SUPPORT-17151?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/input-mapping/src/File/Strategy/AbstractStrategy.php
+++ b/libs/input-mapping/src/File/Strategy/AbstractStrategy.php
@@ -6,12 +6,14 @@ namespace Keboola\InputMapping\File\Strategy;
 
 use Keboola\InputMapping\Exception\FileNotFoundException;
 use Keboola\InputMapping\Exception\InputOperationException;
+use Keboola\InputMapping\Exception\InvalidInputException;
 use Keboola\InputMapping\File\StrategyInterface;
 use Keboola\InputMapping\Helper\ManifestCreator;
 use Keboola\InputMapping\Reader;
 use Keboola\InputMapping\State\InputFileStateList;
 use Keboola\StagingProvider\Staging\File\FileFormat;
 use Keboola\StagingProvider\Staging\File\FileStagingInterface;
+use Keboola\StorageApi\ClientException;
 use Keboola\StorageApi\Options\GetFileOptions;
 use Keboola\StorageApiBranch\ClientWrapper;
 use Psr\Log\LoggerInterface;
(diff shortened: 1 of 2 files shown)


ondrajodas (approved):
ok

PR #557, opened 2026-08-07
docs(libs): AJDA-3033 move agent guidance to AGENTS.md and point PRs at the monorepo

odinuv:
## Link to issue

https://linear.app/keboola/issue/AJDA-3033/popsat-output-mapping-pro-agenty

## Description

Renames the root `CLAUDE.md` and all 23 `libs/*/CLAUDE.md` files to `AGENTS.md` (via `git mv`, so history follows) and leaves a short `CLAUDE.md` pointer next to each. The guidance content is unchanged apart from cross-references; the point of the rename is that `AGENTS.md` is the name Codex, Cursor, Gemini CLI and others look for, while the pointer keeps Claude Code working.

The substantive addition is a **"Contributing — this repository is a mirror"** section at the top of every `libs/<lib>/AGENTS.md`. It states that the standalone `keboola/<repo>` repository is a read-only mirror that CI force-pushes on every green build, and that pull requests must go to `keboola/platform-libraries` instead. It also gives an agent a way to detect where it is ("no `libs/` directory at the root ⇒ you are in the mirror"), plus the commit-scope and `<lib>/<version>` release-tag conventions.

That section is deliberately self-contained, with no `../..` references: after the CI split it becomes the *root* `AGENTS.md` of the standalone repo, and it is the only guidance file that ever reaches an agent working there. The 5 libraries whose standalone repo name differs from the directory name (e.g. `key-generator` → `keboola/php-key-generator`) name their real repo, taken from the `split-library` action's `case`.

Supporting changes: root `AGENTS.md` documents the file layout and why the mirror section must be preserved; the `adding-a-library` skill now requires both files for a new library; `README.md` and the skill follow the rename.

## Justification

An agent given a task in one of the split repos has no way to know it is a mirror — the checkout looks like a normal library repo. https://github.com/keboola/input-mapping/pull/2 is the result: a PR against a read-only repo that can never be merged. Putting the rule in the one file that ships into the mirror is the only place it can actually be read.

None.

## Impact Analysis

Documentation only — no source, test or CI configuration is changed, and **no library is released**. Nothing that consumers install changes.

The one operational side effect is that touching all 23 library directories makes this push run the full CI fan-out (including the Storage-backed input/output-mapping suites). CI also syncs the branch to the standalone repos as it does for every branch; that is a branch mirror, not a release. Once merged, each mirror's `main` picks up an `AGENTS.md` at its root plus the `CLAUDE.md` pointer, which is the intended outcome.

## Deployment Plan

**Merging to `main` is enough — no release.** No version tag will be cut for any library, so no new package version is published.

## Rollback Plan

Revert this PR.

## Post-Release Support Plan

None.


MiroCillik (approved):

ondrajodas (approved):
:+1:

PR #558, opened 2026-08-12
fix(service-client): point Service::OAUTH at oauth-service, not legacy oauth-api

odinuv:
[link to issue](https://linear.app/keboola/issue/AJDA-3069/slsp-github-integration)

## Description

`Service::OAUTH`'s internal name still pointed at `oauth-api.default`, the legacy Node.js service. The PHP `oauth-service` that replaced it is deployed as `oauth-service-api` — the enum was simply never updated with the migration.

```
- self::OAUTH => 'oauth-api.default',
+ self::OAUTH => 'oauth-service-api.default',
```

Only `INTERNAL` DNS is affected; `PUBLIC` goes through the ingress and is unchanged, which is why nobody noticed. Surfaced while wiring sandboxes-service to oauth-service's new GitHub credential endpoints, which exist only on the PHP service.

## Release Notes

- **Justification**
    - In-cluster callers should reach the service that actually implements the API.

- **Plans for Customer Communication**
    - None.

- **Impact Analysis**
    - In-cluster OAuth traffic moves from `oauth-api` to `oauth-service`. I found no consumer using `Service::OAUTH` with `INTERNAL` DNS other than the new sandboxes-service call, but worth a look from whoever owns the migration in case something is deliberately pinned to the legacy service.
    - No change for `PUBLIC` DNS consumers.

- **Deployment Plan**
    - Release a tagged version.

- **Rollback Plan**
    - Revert this PR.

- **Post-Release Support Plan**
    - None.

### DIFF
--- a/libs/service-client/src/Service.php
+++ b/libs/service-client/src/Service.php
@@ -63,7 +63,7 @@ enum Service
             self::GIT_SERVICE => 'git-service.git-service',
             self::IMPORT => 'sapi-importer.default',
             self::NOTIFICATION => 'notification-api.default',
-            self::OAUTH => 'oauth-api.default',
+            self::OAUTH => 'oauth-service-api.default',
             self::QUERY => 'query-service-api.query-service',
             self::QUEUE => 'job-queue-api.default',
             self::QUEUE_INTERNAL_API => 'job-queue-internal-api.default',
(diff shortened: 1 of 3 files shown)


janvanicek (approved):
Lgtm

PR #559, opened 2026-08-12
feat(ci): AJDA-3137 create GitHub releases on the split library repos

odinuv:
**Linear**: [AJDA-3137](https://linear.app/keboola/issue/AJDA-3137/autorelease-in-platform-librarries)

## Description

Pushing a `<lib>/X.Y.Z` tag already splits the tag into the library's standalone repo, but nobody
created a GitHub **release** there — so release notes were written by hand and regularly forgotten.
`keboola/output-mapping` is tagged `27.10.0` with no release for it.

The tag push now creates the release too, with notes written by Claude Code. The split is applied
first and the release second, so the tag exists before anything tries to attach a release to it.

**What generates the notes.** The deterministic half is shell (`bin/release-context.sh`): the
previous released version, the commits in that range touching `libs/<lib>`, and the pull requests
those commits came from. Pull requests are resolved per commit through
`gh api repos/<monorepo>/commits/<sha>/pulls` rather than by parsing merge commits — a release range
also contains merges belonging to *other* libraries, and a path-filtered `git log` drops merge
commits entirely. On `output-mapping/27.9.0..27.10.0` that yields exactly #550, #551, #553 and #556,
excluding #554 (api-bundle) and #555 (input-mapping).

The judgement half is `anthropics/claude-code-action@v1`, which reads those pull request
descriptions — the richest account of what changed — and condenses them into a handful of bullets in
the style of the existing notes. The `**Full Changelog**` compare link is appended in shell so it
can never be wrong.

**Failure behaviour is deliberately loud.** Empty notes fail the job rather than falling back to a
commit dump: a wrong release note is worse than a missing one. Recovery is a run of the new
`create-release.yml` (`workflow_dispatch`), which is also how a version tagged before this existed
gets its release. Both callers skip when the release is already there, so both are safe to re-run.

Only strict `<lib>/X.Y.Z` tags get a release — `<lib>/build-*` and `<lib>/dev-*` tags keep being
published without one.

**Two incidental changes.** The 5 dir≠repo exceptions moved out of `split-library`'s inline `case`
into `bin/target-repo.sh`, because both publishing actions need the mapping now and one source of
truth beats two. And both actions take their credentials from job-level `env` instead of inputs
(`SPLIT_APP_ID`, `SPLIT_APP_PRIVATE_KEY`, `ANTHROPIC_API_KEY`); `claude-code-action` picks the
Anthropic key up on its own via `inputs.anthropic_api_key || env.ANTHROPIC_API_KEY`.

No new GitHub App is needed: creating a release wants the same `contents: write` the publish App
already has to push tags. `AI_CR_ANTHROPIC_API_KEY` is an existing **organization** secret, already
used by the Claude Code workflows in other keboola repos.

## Justification

See the issue — cutting releases by hand is tedious and keeps being skipped, which leaves consumers
of these libraries without a changelog for versions they are asked to upgrade to.

None. Internal CI change; the only externally visible effect is that the read-only library repos
gain release notes.

## Impact Analysis

- **Publishing (all 22 libraries)** is the real risk, not the new release step: `split-library` now
  reads its credentials from `env` and resolves the target repo through `bin/target-repo.sh`. If
  either were wrong, publishing breaks everywhere. This branch's own CI run exercises it — a
  `.github/**` change makes the affected-libraries gate fall back to all 22 libraries, so `publish`
  runs the reworked action for every one of them.
- **A failed notes generation cannot lose a release.** The split runs first, so the tag is already
  published; only the release is missing, and `create-release.yml` creates it afterwards.
- **Job-level `env` is inherited by the Claude Code subprocess**, `SPLIT_APP_PRIVATE_KEY` included.
  That is unavoidable once credentials come from `env` — the token-minting step can only read them
  through `${{ env.* }}`. Compensated with a narrow `--allowedTools`: `gh pr view/diff/api`,
  `git log/show/diff`, `Read`/`Grep`/`Glob`/`Write`, and no general shell that could print the
  environment. Noted in `AGENTS.md` so nobody widens it unaware.
- **Cost**: one Claude Code run per released version, reading a handful of pull requests.
- Not gated by a feature flag. No single-tenant impact — nothing in `libs/` changed.

## Deployment Plan

Merge. Workflows take effect from `main`; no library release is involved, since no library code
changed.

Then, to verify end to end and backfill the release that is actually missing: run **Create release
(manual)** for `output-mapping` `27.10.0` with `dry-run` on, check the notes in the job summary, and
re-run with `dry-run` off. (`workflow_dispatch` only becomes available once the file is on `main`,
which is why this cannot be dry-run from the branch.)

## Rollback Plan

Revert this PR.

## Post-Release Support Plan

None.


pepamartinec (commented):
Superseded by the Request changes review below — same findings, correct state.

.github/actions/create-release/action.yml
```diff
@@ -0,0 +1,210 @@
...
+        NOTES_FILE: ${{ github.workspace }}/.release/notes.md
+        LIBRARY: ${{ inputs.library }}
+      run: |
+        printf '%s\n' \
+          "No changes in \`libs/${LIBRARY}\` itself. Released to pick up changes in the libraries it depends on." \
+          > "${NOTES_FILE}"
+
+    - name: Write release notes with Claude Code
+      if: steps.existing.outputs.skip != 'true' && steps.context.outputs.has-changes == 'true'
+      uses: anthropics/claude-code-action@v1
```
pepamartinec, line 104: **Blocking: this step has no GitHub token and fails before Claude runs.**

`setupGitHubToken()` (`src/github/token.ts`) resolves exactly two ways — `OVERRIDE_GITHUB_TOKEN` (the `github_token` input) or OIDC. There is no third branch; `DEFAULT_WORKFLOW_TOKEN: ${{ github.token }}` exists in the action's `env:` but is only consumed by `src/mcp/install-mcp-server.ts` for the MCP server, not as a fallback.

This PR passes no `github_token`, and `release.yml` grants no `id-token: write`. So `getOidcToken()` throws and the step fails in phase 1 "Prepare" — on every release.

Fix with the input, not the permission:

```yaml
      uses: anthropics/claude-code-action@v1
      with:
        github_token: ${{ github.token }}
        anthropic_api_key: ${{ inputs.anthropic-api-key }}
```

Adding `id-token: write` instead would route auth through the Claude GitHub App — an unstated new org prerequisite, and `DEFAULT_PERMISSIONS` there is `contents: write` + `pull_requests: write` + `issues: write` (`token.ts:67`), strictly more privilege than this PR is trying to hold.

Side note: the `GH_TOKEN` env below is dead either way — `src/entrypoints/run.ts:188` overwrites it with whatever `setupGitHubToken()` returned. Passing `github_token` makes the two agree.

.github/actions/create-release/action.yml
```diff
@@ -0,0 +1,210 @@
...
+      env:
+        GH_TOKEN: ${{ github.token }}
+      with:
+        anthropic_api_key: ${{ inputs.anthropic-api-key }}
+        show_full_output: true
```
pepamartinec: `keboola/platform-libraries` is public, so these logs are world-readable. The action's own warning on this input: *"outputs ALL Claude messages including tool execution results which may contain secrets... Only enable for debugging in non-sensitive environments."*

Not an actual leak — the inputs are already public and both keys are masked secrets — but gratuitous, and the notes already land in `$GITHUB_STEP_SUMMARY`, which is what you want to read. Suggest dropping it.

.github/actions/create-release/action.yml
```diff
@@ -0,0 +1,210 @@
...
+      shell: bash
+      env:
+        GH_TOKEN: ${{ steps.app-token.outputs.token }}
+        NOTES_FILE: ${{ github.workspace }}/.release/notes.md
+        REPO: keboola/${{ steps.target.outputs.repo }}
+        VERSION: ${{ inputs.version }}
+      run: |
+        # --verify-tag runs seconds after split-repo.sh pushed that tag. If this ever fails claiming
+        # the tag does not exist, suspect replication lag on the target repo before anything else.
+        gh release create "${VERSION}" \
```
pepamartinec, line 210: **This marks backfilled releases as "Latest", demoting the real one.**

`gh` only sends `make_latest` when `--latest` is explicit (`cli/cli` `create.go:462`), and the REST default is *"true for newly published releases"*. The `--help` text saying "automatic based on date and version" describes `make_latest: legacy`, which is not what gets sent.

Latent today — `output-mapping`'s newest release is `27.9.0`, so the planned `27.10.0` backfill is fine — but `create-release.yml` exists to backfill older versions, which is exactly the case that breaks.

`release-context.sh` already sorts the strict version tags, so emitting `is-newest=true|false` and passing `--latest=${{ steps.context.outputs.is-newest }}` is a cheap fix.

.github/workflows/create-release.yml
```diff
@@ -0,0 +1,61 @@
...
+        type: string
+      dry-run:
+        description: Only print the notes to the job summary, do not create the release
+        required: false
+        default: true
+        type: boolean
+
+permissions:
+  contents: read
+  pull-requests: read
```
pepamartinec, line 25: **Same token defect as `create-release/action.yml` — so the escape hatch is broken too.**

That undercuts *"A failed notes generation cannot lose a release... `create-release.yml` creates it afterwards."* The tag does survive, but this workflow calls the same action with the same missing token. The feature and its fallback are both non-functional; fix both callers together.

Concretely: the stated Deployment Plan (dry-run `create-release.yml` for `output-mapping` `27.10.0`) will fail on the first attempt.

pepamartinec (changes requested):
Blocking findings are inline. Three that don't attach to a line:

**CI has never exercised the reworked `split-library`.** The Impact Analysis claims `publish` runs the reworked action for all 22 libraries on this branch. It didn't — `publish` was skipped or cancelled on all four runs, because `tests-result` is red from an unrelated Storage API flake in `output-mapping / Slice feature`. That leaves the riskiest change in the PR — all 22 libraries' publishing rerouted through a brand-new `bin/target-repo.sh` — completely untested. Re-run to green before merging.

**Confirm `AI_CR_ANTHROPIC_API_KEY` actually reaches this repo.** Org secrets can be scoped to a repo subset; an unset one becomes an empty string and surfaces as a confusing auth error rather than "secret missing".

**The PR description is stale.** The "both actions take their credentials from job-level `env`" paragraph and the matching Impact Analysis bullet describe code that `c467c54d` reverted — both actions still use inputs and `ci.yml` is untouched, so the `--allowedTools` justification now rests on a premise that isn't there. Matters more than usual here, since PR bodies are this feature's own input.

pepamartinec:
High-level jsem za :+1: Claude nasel nejaky potencialni problemy, tak na to kdyztak jeste mrkni.

MiroCillik (commented):
Code review of the release-notes automation. Four findings inline (1 high, 2 medium, 1 low), plus two notes:

- **The PR description is stale relative to the head commit.** It says "both actions take their credentials from job-level `env` instead of inputs" and analyses the resulting `SPLIT_APP_PRIVATE_KEY` exposure to the Claude subprocess. The code does the opposite — credentials are action inputs (`create-release/action.yml:7-30`), which is the safer design and is what `AGENTS.md:236` now documents. Worth updating so reviewers assess the right design.
- **The deterministic half verifies clean against the real repo.** Previous-tag resolution (strict `X.Y.Z` grep → `sort -V` → `awk` predecessor) is correct across the full tag history for output-mapping / input-mapping / k8s-client, including the 27.9.0 → 27.10.0 case; all 38 non-strict tags in the repo are genuinely build/dev tags; the per-commit PR lookup on `27.9.0..27.10.0` reproduces exactly #550, #551, #553, #556. `--allowedTools` inside `claude_args` is parsed and merged correctly by the action (`parse-sdk-options.ts` escapes `()` before `shell-quote`), `permissionMode` is left at `default` so the list really is enforced, and no tracked `.claude/settings.json` exists that could widen it.

One trivial thing not worth its own comment: the emptiness guard is `[ ! -s "${NOTES_FILE}" ]`, so a notes file containing only a newline would pass and ship.

.github/actions/create-release/action.yml
MiroCillik, line 104: **HIGH — this step will fail on every release: no `github_token` input and no `id-token: write`.**

`claude-code-action@v1`'s `run.ts` calls `setupGitHubToken()` unconditionally, before any mode handling. It returns `OVERRIDE_GITHUB_TOKEN` (the action's `github_token` input) if set; otherwise it calls `core.getIDToken()` and exchanges the OIDC token at `api.anthropic.com/api/github/github-app-token-exchange`. This step passes only `anthropic_api_key`, and neither `release.yml` (lines 6-9) nor `create-release.yml` (lines 23-25) grants `id-token: write` — so `ACTIONS_ID_TOKEN_REQUEST_URL` is unset and `getIDToken()` throws, after retries, with "Could not fetch an OIDC token. Did you remember to add `id-token: write`…".

Concrete scenario: push `output-mapping/27.11.0` → the split succeeds and the tag lands in the standalone repo → this step dies → `Finalize release notes` never runs → job red, no release. Every time there are library changes.

This can't be caught by the branch's own CI, since `release.yml` only runs on tag push and `workflow_dispatch` needs `main`. The sibling workflow already using this secret, `keboola/go-monorepo/.github/workflows/claude-code-review.yml`, grants `id-token: write` for exactly this reason.

Preferred fix is `github_token: ${{ github.token }}` rather than granting `id-token: write` — see the comment on line 99 for why.

.github/actions/create-release/action.yml
```diff
@@ -0,0 +1,210 @@
...
+      run: |
+        printf '%s\n' \
+          "No changes in \`libs/${LIBRARY}\` itself. Released to pick up changes in the libraries it depends on." \
+          > "${NOTES_FILE}"
+
+    - name: Write release notes with Claude Code
+      if: steps.existing.outputs.skip != 'true' && steps.context.outputs.has-changes == 'true'
+      uses: anthropics/claude-code-action@v1
+      env:
+        GH_TOKEN: ${{ github.token }}
```
MiroCillik: **MEDIUM — this `env:` does not control the token the model gets, and the obvious fix to the line-97 failure would hand the session a write-capable token.**

`run.ts` does `process.env.GH_TOKEN = githubToken` using the token from `setupGitHubToken()` before spawning Claude, so this `env:` value is overwritten.

If the missing-token failure is fixed by adding `id-token: write` instead of by passing `github_token`, the exchanged Claude App token is minted with `DEFAULT_PERMISSIONS = { contents: write, pull_requests: write, issues: write }` (`src/github/token.ts`). The model's `gh`/`git` calls would then run under a write-capable token while it is reading attacker-influenced PR bodies — directly contradicting the boundary documented in `AGENTS.md` ("Everything allowed is either local … or a read-only `gh pr` subcommand", "The residual risk is bad *note text*").

Passing `github_token: ${{ github.token }}` fixes both findings and keeps the session on the declared `contents: read` + `pull-requests: read`.

.github/actions/create-release/action.yml
MiroCillik: **MEDIUM — this publishes the whole Claude session to public logs.**

`keboola/platform-libraries` is a public repo, so Actions logs are world-readable. The action's own input description:

> Show full JSON output from Claude Code. **WARNING:** This outputs ALL Claude messages including tool execution results which may contain secrets, API keys, or other sensitive information. These logs are publicly visible in GitHub Actions. Only enable for debugging in non-sensitive environments.

The release notes are already surfaced deliberately — `Finalize release notes` writes them to `$GITHUB_STEP_SUMMARY` — so this looks like leftover debugging with no upside. As written, every release dumps every tool result (full PR bodies, file reads, diffs) into a public log, and anything GitHub's secret masker does not recognise is published.

Suggest dropping the input.

AGENTS.md
```diff
@@ -222,8 +222,10 @@ CI runs on **GitHub Actions**: test the affected libraries on every push to any
...
+- **GitHub releases** — `.github/actions/create-release`, called from `release.yml` after the split and from `create-release.yml` (`workflow_dispatch`: `library`, `version`, `dry-run` defaulting to **true**) for retries and for backfilling versions tagged before this existed. It skips if the release already exists, so both callers are safe to re-run. `bin/release-context.sh` gathers the deterministic part — previous released version, the commits in that range touching `libs/<lib>`, and the pull requests those commits came from — then `anthropics/claude-code-action@v1` turns the **pull request descriptions** into the notes and the action appends the `**Full Changelog**` compare link in shell. **Nothing degrades quietly:** empty notes fail the job, and so do an empty commit list or a failed pull request lookup in `bin/release-context.sh` — a context file with nothing in it would still yield *some* text, which would pass the emptiness check and ship. A wrong release note is worse than a missing one; recovery is a `create-release.yml` run. Only strict `<lib>/X.Y.Z` tags get a release — `<lib>/build-*` and `<lib>/dev-*` tags are still published, just without one.
```
MiroCillik: **LOW — this documents a failure mode the code deliberately does not have.**

> **Nothing degrades quietly:** empty notes fail the job, and so do an empty commit list or a failed pull request lookup in `bin/release-context.sh` — a context file with nothing in it would still yield *some* text, which would pass the emptiness check and ship.

`bin/release-context.sh` does the opposite for an empty commit list: it writes a context file, prints `has-changes=false` and exits 0, and the action then ships the canned "No changes in `libs/<lib>` itself. Released to pick up changes in the libraries it depends on." note. (A failed `gh api` lookup *is* fatal, as documented — that half is accurate.)

The `has-changes` path — the one that fires on the ~14/518 dependency-only re-releases the script's own comment cites — isn't documented at all, so the next maintainer who sees that canned note on a green job will go looking for a bug that isn't there.

.github/workflows/create-release.yml
```diff
@@ -0,0 +1,57 @@
...
+          VERSION: ${{ inputs.version }}
+        run: |
+          if [ ! -d "libs/${LIBRARY}" ]; then
+            echo "::error::'libs/${LIBRARY}' is not a library directory."
+            exit 1
+          fi
+          if ! printf '%s' "${VERSION}" | grep -qE '^[0-9]+\.[0-9]+\.[0-9]+$'; then
+            echo "::error::'${VERSION}' is not a version. Pass it without the library prefix, e.g. 27.10.0."
+            exit 1
+          fi
```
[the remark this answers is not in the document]
odinuv: Fixed in ada7b93c — the library name is checked against `^[a-z0-9]+(-[a-z0-9]+)*$` before the `libs/` existence check, and the version against `^[0-9]+\.[0-9]+\.[0-9]+$`.

.github/actions/create-release/action.yml
odinuv: You were right — fixed in 5d69b2a3 with `github_token: ${{ github.token }}` rather than `id-token: write`, for exactly the privilege reason you give, and the dead `GH_TOKEN` env is gone.

.github/actions/create-release/action.yml
odinuv: Dropped in 5d69b2a3.

.github/actions/create-release/action.yml
odinuv: Good catch — 5d69b2a3 has `release-context.sh` emit `is-newest` and passes `--latest=${{ steps.context.outputs.is-newest }}`, verified `false` on an older tag and `true` on the newest.

.github/workflows/create-release.yml
odinuv: Same action, so the same fix covers it — 5d69b2a3 makes the escape hatch work, and the `27.10.0` dry run is still the first end-to-end proof since neither caller can run from a branch.

.github/actions/create-release/action.yml
odinuv: Confirmed and fixed in 5d69b2a3 via the `github_token` input, keeping the session off the Claude App token as you and pepamartinec both recommended.

.github/actions/create-release/action.yml
odinuv: Removed the dead `env:` block and passed `github_token: ${{ github.token }}` in 5d69b2a3, so the model runs on the caller's `contents: read` + `pull-requests: read` and the `AGENTS.md` boundary holds.

.github/actions/create-release/action.yml
odinuv: Dropped in 5d69b2a3.

AGENTS.md
odinuv: Rewritten in 5d69b2a3 — `AGENTS.md` now documents the `has-changes=false` path and its canned note as a green outcome on the ~14-in-518 dependency-only re-releases, and only claims the genuinely fatal cases.

